PR #223, opened 2013-07-19
Fixed: path traversal attack

gro-ove:


calzoneman:
I'll merge this because adding more sanity checks is never a bad idea, but I'm curious what you were actually able to exploit with this.

What I'm seeing locally is that express automatically rewrites requests for `http://domain/../somefile` to `http://domain/somefile`.

calzoneman:
My fault for not testing this, but it broke requests for `/` sending index.html.  I'll try to be better about testing things, but please also test them before submitting a pull request.

Fixed by https://github.com/calzoneman/sync/commit/d56d46c608b9b88bd3817eae43ba3f4b3ccd50d1

coconono:
can you explain a little what vulnerability we're looking at?

calzoneman:
@coconono I believe @gro-I've was concerned that blindly copying paths would result in a malicious user being able to escape the www folder and read arbitrary files from the server. As far as I can tell, express prevents this by itself, but nevertheless it's probably a good idea to explicitly sanitize it instead of trusting whatever comes from express.

calzoneman:
@gro-ove sorry my phone auto corrected to I've

gro-ove:
Sorry, guys, for I didn't test it. (In place, where I am, is a very bad internet connection, so I couldn't download required libraries.)

My friend asked me to fix it (with this link: http://cytu.be/%252e%252e/%252e%252e/%252e%252e/%252e%252e/%252e%252e/%252e%252e/%252e%252e/%252e%252e/etc/passwd).

Sorry again for that bug and no testing. I promise I will be more responsible in future.

calzoneman:
Ah I see. I guess express doesn't call decodeURIComponent before sanitizing or something.

calzoneman:
Also important to note your fix actually doesn't solve the case of that URL because my server is fully up to date and that worked.

calzoneman:
Actually this appears to be a tricky corner case with the %25 entity (which corresponds to a % sign and thus escapes the URI escape)

calzoneman:
@coconono pull the latest commit and restart immediately

PR #281, opened 2013-09-19
Add /clean and /cleantitle commands for moderators

unbibium:
A tool that can be used against spam and trolls.


calzoneman:
Couple of issues here
1. Merge conflicts due to your branch being behind the main master.  Not sure how hard this will be to merge
2. Your `user_input_re` doesn't seem to work, unless I'm misunderstanding it.  I tried `/cleantitle some title in all lowercase-i` and it didn't match, whereas `/cleantitle Some Title In Correct Case` worked.
3. I think `cleanPlaylist` belongs as a property of a `Playlist` object.  I like the way you designed it to take a filter so it would be reusable.  For this reason, I think it would be better to have the chat handlers just take care of parsing input and checking permissions, then calling `chan.playlist.clean(filter)`.  I wouldn't mind making the change myself if the other issues are addressed.

If these issues are addressed I think it could be a useful feature to have around.

unbibium:
Try `/cleantitle -i some title in all lowercase`

i'll see if i can merge forward...

calzoneman:
Ah I see, I was misunderstanding the syntax of the command.

calzoneman:
@unbibium If you wouldn't mind, could you document these commands on https://github.com/calzoneman/sync/wiki/Beginner%27s-Guide-and-FAQ ?

Thanks

PR #286, opened 2013-09-28
Don't update media if there is no #videowrap

nuclearace:


calzoneman:
Justification?  Is this causing problems for people?  I always got JS errors from the missing DOM elements but it did not affect the functionality of other components.  I'd prefer to not have to add extra checks everywhere (for example, if I ever decided to rename `#videowrap` this would require updating).

nuclearace:
Not causing problems so much as generating the JS errors. Hundreds of them if you're in chat-only for awhile.

PR #326, opened 2014-01-27
Basic theme support

Kirtaner:
Easier extendability for third-party installations without merge conflicts or overwriting files

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -2895,13 +2895,7 @@ Channel.prototype.handleChat = function (user, data) {
         user.socket.emit("chatCooldown", 1000 / this.opts.chat_antiflood_params.sustained);
     }
 
-    if (muted) {
-        user.socket.emit("noflood", {
-            action: "chat",
-            msg: "You have been muted on this channel."
-        });
-        return;
-    } else if (smuted) {
+    if (smuted) {
         msg = XSS.sanitizeText(msg);
         msg = this.filterMessage(msg);
         var msgobj = {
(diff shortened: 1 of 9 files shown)


calzoneman:
I like the idea.  It reminds me of the way themes used to work on the discussion board software I used.  I only have one concern: user selected themes from the Options menu.  Would you require every theme to override default/dark/whatever other theme (I plan to add a dark colorscheme, just haven't gotten around to it)?  What happens if the server owner switches the theme path, and therefore everyone's `USEROPTS.theme` points to the wrong directory?

calzoneman:
I'd love to merge this in, I just want to make sure these points are addressed first.

calzoneman:
@Kirtaner

Kirtaner:
Bit busy lately. Anyways, there's a few ways we can handle this. 

Method 1 - USEROPTS is theme dependant and resets to defaults on theme changes. Considering this is unlikely to occur on a regular basis, that's one approach. 

You'd do something like setting USEROPTS.sitetheme = 'cytube', and on pageload if the stored value isn't matching up with what's on the server, reset it all.

Method 2 - A list of supported settings values in the config. Anything outside these values is dropped. Probably not the best choice.

calzoneman:
Well, more of what I was going for is where does the separation of user-chosen CSS and the site theme lie in your proposed idea?

Currently `USEROPTS.theme` links to a CSS file.  If the jade templates on the server are not the default theme, this CSS might have unintended side effects.  Currently the CSS choices are hardcoded in, I suppose I would have to change that to dynamically see which CSS colors are available in the currently selected server theme.

Kirtaner:
Well, if it helps to understand, in my current use case, `USEROPTS.theme` isn't even used and the corresponding setting dropdown is removed.

I suppose a list of user themes available to the server theme can be defined like so

`html-template:
  theme: 'cytube',
  userthemes: 
    - 'default'
    - 'semidark'
    - 'dark'
  title: 'CyTube',
  description: 'Free, open source synchtube'`

If only one usertheme is defined, don't display the option. If there's a mismatch, reset `USEROPTS.theme`

This has an added benefit of allowing easier extra user themes without template editing.

PR #433, opened 2015-02-02
CSS Cleanup + Modern theme

gv1222:
I've spent some time cleaning up the CSS and creating a new theme. I apologize for the messiness of the PR 
Screenshot of modern theme: 

image


www/css/cytube.css
```diff
@@ -107,13 +96,6 @@
     cursor: pointer;
 }
 
-#userlist, #messagebuffer {
```
calzoneman, line 110: Where did these properties get moved to?

www/css/cytube.css
```diff
@@ -173,24 +155,14 @@
...
-#plmeta {
-    border: 1px solid #aaaaaa;
-}
+
 
 #pllength {
     float: right;
 }
 
-.queue_entry {
```
calzoneman, line 184: Is there any particular reason to get rid of the default properties (e.g. `line-height`, `padding`, `font-size`)?  Seems like the only thing that should be overridden by the theme in this case is the border color.

www/css/cytube.css
```diff
@@ -266,16 +238,24 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
...
 .server-msg-disconnect {
     border: 1px solid #ff0000;
-    margin: 5px;
-    padding: 5px;
+    padding-top: 5px;
+    padding-bottom: 5px;
+    margin-top: 5px;
+    margin-right: 5px;
     color: #ff0000;
+    text-align: center;
```
calzoneman, line 246: This CSS class is also used for kick/ban messages as well as the message that you can't chat because you're muted.  Have you checked to see whether `text-align: center` makes those look bad?

calzoneman:
GitHub refuses to show the diffs of slate and cyborg, but as mentioned in IRC please fix the `.drink` border color on slate and cyborg.

www/css/cytube.css
```diff
@@ -266,16 +238,24 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
...
+    background-color: #352727;
```
calzoneman: This background color looks awful on the light/bootstrap themes.  Can this be defined in each theme instead?

www/css/cytube.css
```diff
@@ -266,16 +238,24 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
...
     border: 1px solid #009900;
-    margin: 5px;
-    padding: 5px;
+    padding-top: 5px;
+    padding-bottom: 5px;
+    margin-top: 5px;
+    margin-right: 5px;
     color: #009900;
+    text-align: center;
+    background-color: #273529;
```
calzoneman: Same as above note about `-disconnect`

www/css/cytube.css
calzoneman, line 246: I just checked kick/mute messages, and the centering looks fine as long as the background colors are fixed.

calzoneman:
The foreground color for dropdowns on the modern theme needs to be lightened:

2015-02-02-183518_889x129_scrot

templates/channel.jade
```diff
@@ -38,7 +38,7 @@ html(lang="en")
           #main.row
             #chatwrap.col-lg-5.col-md-5
               #chatheader
-                i#userlisttoggle.glyphicon.glyphicon-chevron-up.pull-left.pointer(title="Show/Hide Userlist")
+                i#userlisttoggle.glyphicon.glyphicon-chevron-right.pull-left.pointer(title="Show/Hide Userlist")
```
calzoneman, line 41: Not opposed to changing this chevron (and maybe I should actually make it switch directions with JS), but can you check how easily it could be aligned better on the other themes?  It looks fine on Modern (maybe because it's far away from the centered text) but on Light it looks strangely misaligned.

www/css/cytube.css
```diff
@@ -173,24 +155,14 @@
...
-    line-height: 22px;
-    padding: 2px;
-    font-size: 8pt;
-    border: 1px solid #aaaaaa;
-    border-top-width: 0;
-}
-
 .queue_temp {
-    background-image: url(/img/stripe-diagonal.png);
+    background-image: url(/img/stripe-diagonal-black.png);
```
calzoneman: Any particular reason for this?  It looks worse on the light themes.

gv1222:
image

www/css/cytube.css
```diff
@@ -266,16 +238,24 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
...
 .server-msg-reconnect {
     border: 1px solid #009900;
-    margin: 5px;
-    padding: 5px;
+    padding-top: 5px;
+    padding-bottom: 5px;
+    margin-top: 5px;
+    margin-right: 5px;
     color: #009900;
+    text-align: center;
```
Xaekai, line 257: Just curious, why isn't this in modern.css so it's specific to your theme?

Wrenuld, line 257: I think it looks quite neat across all themes, if you think it looks bad or have a problem with it I can make it theme specific though. Is there anything specific you don't like about it?

calzoneman, line 257: I don't think there's a problem with centering it everywhere, unless a bunch of people have custom styles that would be broken by it.

Xaekai, line 257: Having tested it with various lengths of kick messages, I guess its ok.

www/css/cytube.css
```diff
@@ -44,14 +43,6 @@
     margin-bottom: 0;
 }
```
gv1222: userlist and message buffer have been moved here and split into different classes

www/css/cytube.css
gv1222, line 110: userlist and message buffer were moved when moving css and are now on line 25-44. My apologies

www/css/cytube.css
gv1222, line 184: This has been resolved and the default properties have been returned to cytube.css

calzoneman:
@gv1222 Can you revert the change to the file permissions on `run.sh`?  It's marked non-executable now.

PR #434, opened 2015-02-05
Userlist toggle changes

Xaekai:
Change chevron appearance on toggle.
Change chevron placement for Synchtube layout

### DIFF
--- a/templates/channel.jade
+++ b/templates/channel.jade
@@ -38,7 +38,7 @@ html(lang="en")
           #main.row
             #chatwrap.col-lg-5.col-md-5
               #chatheader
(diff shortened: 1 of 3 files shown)


calzoneman:
1. The toggle initially has the wrong orientation.  It starts pointing right, then when you click it it still points right, then when you click again it points down
2. Are you sure about using right and down?  It seems a bit odd to me, but I'm not sure if I'd like right/left or up/down any better.

Xaekai:
My logic was:
Down for open userlist
Pointing away from whatever edge the userlist is on for closed.

I will update the channel.jade to account for the fact that the userlist's default state is open.

PR #444, opened 2015-02-15
Refactored all templates to extend from a base layout

LavaToaster:


calzoneman:
I'm not sure that now is the best time to merge this.  I know of at least a few people with local changes to these templates that this would cause merge headaches for, and all of these changes would have to be done over again when 4.0 is released.

I think the better route would be for me to make my development work on 4.0 public and draft some guidelines for contributing, then allow you to make a similar patch there.  4.0 is nowhere near ready for use yet, but for that same reason I'm more willing to make these kinds of changes to it, since nobody is using it and it won't break for anyone.

LavaToaster:
That's cool. 

If 4.0 is a rewrite and how far you've come along with it, it shouldn't be that hard to adopt this style from the beginning.

PR #446, opened 2015-02-16
Enhance media link parser.

### LINKED ISSUE #445: Dailymotion duping
Test case:
Channel options: Allow Duplicate Videos unchecked.
Queue: www.dailymotion.com/video/x2geddf_a-history-of-lego-in-three-minutes_tech
Queue: www.dailymotion.com/video/x2geddf
Queue: www.dailymotion.com/video/x2geddf_you_can_put_whatever_you_want_here_dont_matter

Xaekai:
Accept the shorthand URI style used in the logs as valid.
Add an underscore to the DailyMotion negated group to prevent dupe abuse. (Resolves #445)

### DIFF
--- a/lib/get-info.js
+++ b/lib/get-info.js
@@ -418,7 +418,7 @@ var Getters = {
     dm: function (id, callback) {
         var m = id.match(/([\w-]+)/);
         if (m) {
-            id = m[1];
+            id = m[1].split("_")[0];
         } else {
             callback("Invalid ID", null);
             return;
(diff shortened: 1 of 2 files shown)


calzoneman:
For the shorthand adds, why not just have a generic matcher like

``` javascript
if ((m = url.match(/([a-z]{2}):([^\?&#]+)/)) {
    return {
            id: m[2],
            type: m[1]
    };
}
```

instead of copy/pasting the same code for every kind of provider?

PR #450, opened 2015-02-21
Drop root privledges

ghost:
I have added a feature allowing users to change uid and gid after starting as root. this is particulary handy for binding ports below 1024.

### DIFF
--- a/lib/setuid.js
+++ b/lib/setuid.js
@@ -11,5 +11,5 @@ if (Config.get("setuid.enabled")) {
             console.log('Cowardly refusing to keep the process alive as root.');
             process.exit(1);
         }
-    }, 15);
+    }, (Config.get("setuid.timeout")));
 };
--- a/lib/setuid.js
+++ b/lib/setuid.js
@@ -1,0 +1,15 @@
+var Config = require("./config");
+
+if (Config.get("setuid.enabled")) {
+    setTimeout(function() {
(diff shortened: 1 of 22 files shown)


lib/setuid.js
```diff
@@ -0,0 +1,15 @@
+var Config = require("./config");
+
+if (Config.get("setuid.enabled")) {
+    setTimeout(function() {
```
calzoneman, line 4: Why is this in a setTimeout?

ghost, line 4: because if it isn't it drops the uid before the ports are bound

calzoneman, line 4: How did you determine the timeout?  My concern would be that if the port bind happened to take uncharacteristically long, it could be a tricky nondeterministic issue to debug.

The ideal solution would be to add the callback to `http.createServer` etc. so that the setuid happens after the port finishes binding.  I'm not sure if there's a clean way to do this for multiple servers, besides keeping track of a counter variable or something.

ghost, line 4: I determined the timeout by tring it on my old 1.6ghz single thread laptop. I think 15ms should be long enough to bind ports. if 15ms is long enough on a 10 year old laptop then I think it should be right.

ghost:
I have made it so it is possible to change the timeout before it drops root.

PR #455, opened 2015-02-25
Update modern theme

gv1222:
This update cleans up the modern theme and fixes some of the inconsistencies. Background images are also better supported with transparency. 

dseq7a78


calzoneman:
Ok, this is where it would be useful to have modern.css not be minified, since I can't even see what half of the changes are.

I'll have to pull this down locally this evening and take a look.

gv1222:
The modern bootstrap is available on github, you can view the changelog here: https://github.com/gv1222/Modern/commits/master. I will unminify the bootstrap later.

calzoneman:
New poll button is still not aligned...
2015-02-26-112644_1215x86_scrot

calzoneman:
The padding on the message box looks a little bit odd to me with nick highlights and hovers.  It doesn't look as bad as last time when it was only applied to the left side, but I dunno.  Might be worth getting more opinions on it.

2015-02-26-112913_490x109_scrot

calzoneman:
A couple other notes:
- Can the script access background be a little more opaque?  It looks a little bit visually noisy with the chat in the background: 
  2015-02-26-113331_200x211_scrot
- The chat header (n connected users) has a semitransparent background, but the video title does not.

gv1222:
I'm not sure what you're seeing on the chatheader, it looks fine on my side. maybe just a cache issue?
6yluctp

www/css/cytube.css
```diff
@@ -239,8 +241,9 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
 .server-msg-disconnect {
     border: 1px solid #ff0000;
     padding-top: 5px;
-    padding-bottom: 5px;
-    margin: 5px;
+    margin-top: 5px;
+    margin-bottom: 5px;
+    height: 30px;
```
calzoneman: Why is `height` set explicitly here?

www/css/cytube.css
```diff
@@ -249,8 +252,9 @@ li.ui-sortable-helper, li.ui-sortable-placeholder + li.queue_entry {
 .server-msg-reconnect {
     border: 1px solid #009900;
     padding-top: 5px;
-    padding-bottom: 5px;
-    margin: 5px;
+    margin-top: 5px;
+    margin-bottom: 5px;
+    height: 30px;
```
calzoneman: Same question as above

www/css/cytube.css
gv1222: The margin-top & margin-bottom are used to keep a gap between the chat and top of message buffer. Without the height of 30px the box height is shrunk from its correct size: 
image. 

Correct/current size
image

calzoneman: I don't think that setting an explicit pixel height for the box is the right way to solve your problem of "correct" size.  Different themes (or channel CSS, or stylish, or whatnot) may use a different font size or line-height, and furthermore in the case of `.server-msg-disconnect`, if someone is on a small screen and receives a long error message, it will be linewrapped, hence breaking the explicit height.

Currently, the height is automatic with `padding-bottom` and `padding-top` being set to prevent the box from being too narrow around the text.  I'm confused why you removed `padding-bottom` and set `height` instead-- is there some reason that the automatic height calculation breaks specifically for Modern?

calzoneman: Essentially what I'm confused about is why you're changing it with the goal of getting the same end result (the current height of the box).  Changing the `margin: 5px` to `margin-top` and `margin-bottom` makes sense, since with the chat messagebuffer padding it's no longer necessary to have left/right margins set on these messages.  However, making those changes only, I don't observe breakage of the height of the Connected box.

PR #457, opened 2015-03-05
set SSL ciphers

ghost:
Have added a way to disable/enable ssl ciphers.
Now sync uses openssl ciphers "HIGH:!DSS:!aNULL@STRENGTH" instead of weaker ciphers like RC4

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -72,7 +72,8 @@ var Server = function () {
         key: key,
         cert: cert,
         passphrase: Config.get("https.passphrase"),
-        ca: ca
+        ca: ca,
+	ciphers: Config.get("https.ciphers")
     };
 
     Config.get("listen").forEach(function (bind) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -73,7 +73,8 @@ var Server = function () {
         cert: cert,
         passphrase: Config.get("https.passphrase"),
         ca: ca,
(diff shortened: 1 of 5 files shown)


calzoneman:
From http://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener:

> Note that it is still possible for a TLS v1.2 client to negotiate a weaker cipher unless `honorCipherOrder` is enabled.

So I think if you're going to change the cipher list you should probably set `honorCipherOrder: true` as well, no?

Getting rid of RC4 is fine by me, but are there any other ciphers your change drops that might cause browser/OS incompatibilities?

ghost:
According to
https://www.ssllabs.com/ssltest/analyze.html?d=synchtu.be&latest the only
browsers this will break is internet explorer 6 on windows xp. I guess I
can look at setting honorCipherOrder: true tomorrow

lib/server.js
```diff
@@ -72,7 +72,8 @@ var Server = function () {
         key: key,
         cert: cert,
         passphrase: Config.get("https.passphrase"),
-        ca: ca
+        ca: ca,
+	ciphers: Config.get("https.ciphers")
```
calzoneman: Please fix indentation

calzoneman:
As long as you've looked into it and determined that what is in config.template.yaml is a sensible default, I'm good to merge this.

PR #463, opened 2015-04-10
Allow canvas elements through the XSS filter

Xaekai:
They were allowed before the upgrade to sanitize-html and pose little threat.


calzoneman:
:why:

Xaekai:
I wanted it for MOTD usage. I don't think it would be useful in messages, haha.

The problem with injecting the element with JS is that any MOTD updates destroy it and the state is lost and everything has to be reinitialized. Just a matter of convenience

PR #466, opened 2015-04-19
Sanitized output of channel name in invalid channel

LavaToaster:
Example request ([link](http://cytu.be/r/%3Cimg%20src=%60%60%20onerror=alert%28%27xxs%27%29%3E))

```
http://cytu.be/r/%3Cimg%20src=%60%60%20onerror=alert%28%27xxs%27%29%3E
```

This does nothing on Chrome as it triggers the XSS auditor, however on Firefox this does trigger the alert.

### DIFF
--- a/lib/web/webserver.js
+++ b/lib/web/webserver.js
@@ -15,6 +15,7 @@ var static = require("serve-static");
 var morgan = require("morgan");
 var session = require("../session");
 var csrf = require("./csrf");
+var XSS = require("../xss");
 
 const LOG_FORMAT = ':real-address - :remote-user [:date] ":method :url HTTP/:http-version" :status :res[content-length] ":referrer" ":user-agent"';
 morgan.token('real-address', function (req) { return req._ip; });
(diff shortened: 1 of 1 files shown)


calzoneman:
I think a better solution would be to call `res.type('text')` rather than invoke the XSS filter.  Not only would that prevent XSS entirely, it also prevents people from embedding legal HTML (e.g. with links to advertisements or something).

calzoneman:
Oh I see, you were calling `sanitizeText` and not `sanitizeHtml`

calzoneman:
Well, the `res.type()` solution is still probably better than calling that, what do you think?

calzoneman:
@Lavoaster also for future reference, XSS is probably a good candidate for PMing me first rather than posting publicly on GitHub.

LavaToaster:
That is a solution. However, Not that it isn't a problem for me, but on EQTV the invalid channel [page](http://equestria.tv/r/%3Cimg%20src=%60%60%20onerror=alert%28%27xxs%27%29%3E) returns something a little more friendlier and that would break it.

LavaToaster:
I did, last week.

calzoneman:
I don't have any record of you telling me about this.  On 2015-04-11 you said "Hey There" but nothing else.  Even if I'm afk, I do always read scrollback as well as my email.

calzoneman:
I'm not mad, just pointing out for the future I would appreciate it.

PR #467, opened 2015-04-20
make sure the user understand this is for real

nuclearace:


www/js/util.js
```diff
@@ -2514,6 +2514,11 @@ function formatUserPlaylistList() {
             .attr("title", "Delete playlist")
             .appendTo(btns)
             .click(function () {
+                var really = confirm("Are you sure you want to delete" +
+                    " this playlist? This cannot be undone.");
+                if (!really) {
```
shlomogoldberg: //you found a shekel

www/js/util.js
```diff
@@ -2514,6 +2514,11 @@ function formatUserPlaylistList() {
...
+                    return;
+                }
                 socket.emit("deletePlaylist", {
                     name: pl.name
                 });
```
shlomogoldberg: add comments you filthy goym

PR #468, opened 2015-04-27
Update chat antiflood limit

calzoneman:
As mentioned previously, allowing channels/moderators to have no limit is not really a good idea, however applying the channel limit to moderators and having a default limit that is relatively slow is not what users want.

The compromise is this:
- Channels must set a burst count <= 20 and a sustained rate <= 10 messages/sec
- If channel rate limiting is disabled, a global limit of 10 messages/sec (with 10 burst) is enforced
- Channel limits are not applied to moderators, instead the global limit from the previous point is used.
- The global limit, not the channel limit, is applied to PMs

### DIFF
--- a/lib/channel/chat.js
+++ b/lib/channel/chat.js
@@ -24,7 +24,7 @@ const TYPE_PM = {
 
 // Limit to 10 messages/sec
 const MIN_ANTIFLOOD = {
-    burst: 10,
+    burst: 20,
     sustained: 10
 };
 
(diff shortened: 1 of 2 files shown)


Xaekai:
So if the channel burst is set above 10, mods are effectively more restricted than regular users?

calzoneman:
I didn't think of that until after I had shut down my computer.  `MIN_ANTIFLOOD` is now exactly the limits of the channel settings.

Xaekai:
I can live with a burst of 20. As a bonus, my $flood command won't be made useless.

calzoneman:
Were there any remaining issues with this in testing or can I merge it?

Xaekai:
I didn't test the PMs but I tested
throttle on + mod | throttle on + user
throttle off + mod | throttle off + user
and everything was in order

Xaekai:
Tested PM's with scripting. It definitely works the way you intended. Though, you can still send quite an excessive amount of spam to someone through PM with scripting, you certainly won't be able to use it as an attack vector.

PR #470, opened 2015-05-06
Fix annoyance: Clicking Remove video scrolls to top

Xaekai:


www/js/util.js
```diff
@@ -1741,6 +1741,7 @@ function removeVideo() {
 
     $("#videowrap").remove();
     $("#chatwrap").removeClass("col-lg-5 col-md-5").addClass("col-md-12");
+    if (event) event.preventDefault();
```
Xaekai, line 1744: The existence check is necessary because removeVideo may be called at pageload for USEROPTS.hidevid, and obviously there is no event to pass in that case.

PR #478, opened 2015-05-19
Fix shadowmuted users usage of greentext and slash commands

### LINKED ISSUE #477: Shadowmute bugs
Both the afk slash command and greentext arrow break for users that are shadowmuted.

This allows shadowmuted users to detect their status as such.

I believe ChatModule.prototype.processChatMsg needs to be slightly reorganized.

Xaekai:
Tested on PonkTube

Closes #477

### DIFF
--- a/lib/channel/chat.js
+++ b/lib/channel/chat.js
@@ -268,6 +268,25 @@ ChatModule.prototype.processChatMsg = function (user, data) {
         return;
     }
 
+    if (data.msg.indexOf(">") === 0) {
+        msgobj.meta.addClass = "greentext";
+    }
+
+    if (data.msg.indexOf("/") === 0) {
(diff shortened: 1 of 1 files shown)


calzoneman:
I haven't tested this, but from a quick read-through it looks like this has the unintended side effect of allowing shadowmuted users to send messages with `/me` and `/sp`.  Since command handlers are responsible for sending their own messages, if you handle all commands before the early exit for shadowmuted users, the messages are broadcast to everyone.

I think what you'll have to do instead is just emulate those commands.

Xaekai:
I included those in my testing and they seemed to work correctly.

Admin view:
image

Testing user:
image

Regular observer:
image

calzoneman:
Oh I see, it has to do with the way `processChatMsg` is called recursively in 3.x

PR #480, opened 2015-05-21
Fix a typo

ghost:
(Also, it might be because I'm using YT's HTML5 thing, but I don't think hidePlayer/unhidePlayer even _do_ anything...)

### DIFF
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -2854,7 +2854,7 @@ function googlePlusSimulator2014(data) {
 
 function EmoteList() {
     this.modal = $("#emotelist");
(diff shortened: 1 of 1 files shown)


calzoneman:
`hidePlayer` and `unhidePlayer` don't do anything unless you have a certain user-agent, because historically the "videos showing over modals" bug was specific to Chrome.  It's not really useful as is, but I need to change those functions to try to intelligently determine whether the modal can be shown over the video.

calzoneman:
However this is a typo so I'll go ahead and merge it.

PR #487, opened 2015-06-07
Remove YouTube v2 API fallback since v2 is dead

calzoneman:
@nuclearace cacn you look this over and merge?


nuclearace:
Looking over

nuclearace:
Looks good to me

PR #489, opened 2015-06-19
Change chat's font to Consolas

dylanrenwick:
This prevents guests "stealing" other users name by replacing lowercase l with uppercase I and vice versa.
Courier is another alternative font that will fix this issue.

### DIFF
--- a/www/css/cytube.css
+++ b/www/css/cytube.css
@@ -42,6 +42,7 @@
     overflow-x: hidden;
     overflow-y: scroll;
     margin-bottom: 0;
+    font-family: Consolas;
 }
 
 #chatline, #guestlogin > input, #guestlogin > .input-group-addon {


calzoneman:
You are more than welcome to apply this to your own channels using CSS, but I don't think that a lot of people will be happy with chat default being changed to a fixed-width font.

calzoneman:
I've been thinking about a real "fix" for this (although it's really an issue of typographical gullibility) which involves just stripping 1, l, and I from usernames before checking for uniqueness.

dylanrenwick:
I was about to suggest other fonts here (such as calibri), but your fix probably works better.

calzoneman:
Changing the font doesn't really solve the problem, since people who don't have that font may end up with one that has the same visual difference issue as the default.

calzoneman:
I've implemented a [fix](https://github.com/calzoneman/sync/commit/c28dc0d3d24360385f0ab9bd9481cde40c3f7aad) in the fix-489 branch, which is also live on http://beta.synchtube.me (but not on cytu.be since it requires a server restart).  You can play around with it and see if it handles the cases you were dealing with.

Note that it does not change the matching mechanism for guest names and other guest names (there can be a `calzoneman` and a `caIzoneman` if both are guests, but not if one of the accounts is registered).  It is only a change to the database matching.

calzoneman:
Moving discussion to #490.

PR #493, opened 2015-07-01
Rewrite video players

calzoneman:


calzoneman:
Known Issues:
- ~~VideoJS seems to get stuck showing the loading spinner after being seeked.  It goes away if you pause and unpause, but perhaps there's a better way?  Another option is to just override the CSS and nuke the spinner entirely.~~ Fixed-- was specific to IE
- ~~Dailymotion possibly has the same initial volume bug that plagued Soundcloud~~ not fixable
- ~~VideoJS volume isn't set correctly~~
- ~~Source type should be `audio/<codec>` for raw audio files~~
- ~~Leader is broken~~
- ~~VideoJS doesn't stop if it transitions to a video that it can't play~~

calzoneman:
Temecula has requested that the livestream.com player be made chromeless again.  I think I'll make it a user option.

PR #512, opened 2015-10-14
fix playlist scrolling performance in chrome

codl:
In chrome (and probably other webkit/blink-based browsers) the scrolling performance of the playlist is awful when there are a lot of items in the playlist and the individual item buttons are displayed. This change forces the playlist to be on its own compositing layer, improving performance significantly.

### DIFF
--- a/www/css/cytube.css
+++ b/www/css/cytube.css
@@ -156,6 +156,7 @@
     list-style: none outside none;
     max-height: 500px;
     overflow-y: auto;
+    transform: translateZ(0);
 }
 
 
--- a/www/css/cytube.css
+++ b/www/css/cytube.css
@@ -151,8 +151,9 @@
 }
 
 .videolist {
+    padding: 0;
+    margin: 0;
     list-style: none outside none;
-    margin-left: 0;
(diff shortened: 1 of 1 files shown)


Xaekai:
This destroys the playlist tooltips.

calzoneman:
I applied this locally to try it out and it seems to cause more problems than it solves.  Scrolling through extremely large playlists is faster, yes, but simply scrolling through the list is not what it's used for.  Reordering playlist items is still very laggy and the item that you grab disappears after this change.  As Xaekai mentioned, this breaks anyone adding tooltips as well.  It seems like kind of a hack that shoves the rendering on the GPU but breaks a lot of CSS behavior in the process.

For better performance on very large playlists in Chrome, you can go to Options, Playback and select "Hide playlist buttons by default".  This will not render the buttons until you right click a playlist item to expand it.  I would also recommend trimming down your playlist, as large playlists are prone to performance issues (even with this patch applied, my tablet is unusable with a 2,000 item playlist).

codl:
Fair enough, I didn't notice that it broke other stuff. Closing this.

PR #569, opened 2016-04-09
Playlist improvements

EusebeG:
Hi
is it possible to have a vote button for each song of the queue? like this example [(https://togethertube.com/rooms/crarmy)]
thx

### DIFF
--- a/lib/database.js
+++ b/lib/database.js
@@ -403,18 +403,7 @@ module.exports.saveUserPlaylist = function (pl, username, plname, callback) {
 
     var tmp = [], time = 0;
     for(var i in pl) {
-        var e = {
-            id: pl[i].media.id,
-            title: pl[i].media.title,
-            seconds: pl[i].media.seconds || 0,
-            type: pl[i].media.type,
-            meta: {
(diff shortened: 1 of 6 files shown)


Xaekai:
Should this be an issue and not a PR?

calzoneman:
I assume you meant to open an issue instead of creating a PR from an outdated branch.

I appreciate your suggestion, but at this time I'm not interested in implementing voting for playlist items.  If your channel needs to be able to vote on videos, you can have moderators open a poll and then add the selected video.

EusebeG:
great thx for quick answer
i find new poll less convenient but i understand your point of view
congratulations for cytu.be

calzoneman:
Sure, I'm not saying that it's a convenient solution, but I was suggesting if you _depend on polling for videos_ then that's the supported way to do it.  CyTube was not designed for the use case of voting for each video, it was designed for users adding videos to a first-in, first-out playlist, optionally skipping especially unlikable videos.

EusebeG:
i see...and i understand. It was only a suggestion.
Thank you for your kindness

PR #587, opened 2016-07-07
Update from Jade to Pug

Xaekai:
As you are aware, Jade has been forced to rename. So may as well get this out of the way.

_Changes:_
- module dependency updated from jade 1.11.0 to pug 2.0.0-beta3
- All references to Jade have been changed to Pug
- /srv/web/jade.js is renamed to pug.js
- all template files renamed accordingly
- "mixin somename" is automatically considered a declaration, invocations must use "+somename"
- variable interpolation is no longer supported inside element attributes, use direct references and string concatenation instead.


package.json
```diff
@@ -2,7 +2,7 @@
   "author": "Calvin Montgomery",
   "name": "CyTube",
   "description": "Online media synchronizer and chat",
-  "version": "3.17.5",
+  "version": "3.17.6",
```
calzoneman: I've been pretty bad about following this in the past, but I think that this particular change is worthy of bumping the minor version to 18.

calzoneman:
Bit of a shame they removed interpolation, that was convenient.  Oh well.  Did you test all the pages to make sure they work?

Xaekai:
I tested this on PonkTube and made sure to test the index, a channel, the user account channels page, deleting a channel, account pass/email page, account profile page, the acp, logging out, logging in, and the contacts page.

I will push an update on the version now.

Xaekai:
Done. I will now adjust the other PRs accordingly, in addition to any other required changes.

PR #588, opened 2016-07-07
Add some various harmless tags to the XSS whitelist

### LINKED ISSUE #579: Allow sup and sub tags
Change the tag filter to allow the `<sup>` and `<sub>` elements for formatting text.

Xaekai:
sub, sup: Closes #579
cite, small: Bootstrap uses these for blockquotes
template: Will allow for cleaner channel scripts. Since it's contents are inert it will also allow channel admins to have "comments" in their banner.


calzoneman:
Lel wut, why is this merge confli- oh because you changed the version number.

Xaekai:
Do I need to take some sort of action? I didn't realize it would create a conflict since I presumed the PR's would be pulled in chronological order, and only the version line was changed.

calzoneman:
So even if they are pulled in order, since the same line was changed in order and they aren't rebased between merges, it requires manual effort to merge.  I'll probably just use the github-provided commands to manually merge and fix the conflicts.

Xaekai:
There.

PR #589, opened 2016-07-07
Add timestamps to polls

### LINKED ISSUE #562: poll timestamps
it would be nice if there was a creation timestamp visible for the polls.

Xaekai:
The use of .data() on the timestamp label is so channel script that modifies polls can access the original timestamp of any preexisting poll before the script has had a chance to run since the chatMsg style formatting is lossy.

Using "time" instead of "timestamp" in packUpdate makes the socket frame consistent with style of others, such as chatMsg. I didn't use "time" everywhere else though, because I didn't want it to get confused with "timer" which is also present.

Closes #562


calzoneman:
I think I'd rather keep the consistency within the poll module rather than consistency with other frames that were written 3 years ago.  If I could go back in time (heh), I would probably make the `chatMsg` key be `timestamp` as well, but it's too late for that.

Xaekai:
Done.

PR #590, opened 2016-07-07
Add missing formatter entry for vid.me

Xaekai:


calzoneman:
I think there's another instance of the formatter somewhere in `src/` but I could be wrong.

Xaekai:
Done.

PR #593, opened 2016-07-08
Default theme reference refactor

Xaekai:
Reduce the number of places the default theme is hardcoded from 6 to 1.

### DIFF
--- a/templates/head.pug
+++ b/templates/head.pug
@@ -1,3 +1,5 @@
+- var DEFAULT_THEME = "/css/themes/slate.css";
+
 mixin head()
   meta(charset="utf-8")
   meta(name="viewport", content="width=device-width, initial-scale=1.0")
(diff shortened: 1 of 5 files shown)


www/js/data.js
```diff
@@ -97,7 +97,7 @@ function getOrDefault(k, def) {
 }
 
 var USEROPTS = {
-    theme                : getOrDefault("theme", "/css/themes/slate.css"),
+    theme                : getOrDefault("theme", DEFAULT_THEME), // Set in head template
```
calzoneman, line 100: Does this actually work?  I thought jade locals were local to the jade renderer.

Xaekai:
The jade local was interpolated into a declaration in a script in the template (line 15 of head.pug). I tested this on PonkTube.

The comments in data and theme are to guide neophyte server operators to the right place to change the default.

The head template seems like the ideal place for that setting to live since it's loaded by every page.

templates/head.pug
```diff
@@ -7,7 +9,10 @@ mixin head()
   title= siteTitle
   link(href="/css/sticky-footer-navbar.css", rel="stylesheet")
   link(href="/css/cytube.css", rel="stylesheet")
-  link(id="usertheme", href="/css/themes/slate.css", rel="stylesheet")
+  link(id="usertheme", href=DEFAULT_THEME, rel="stylesheet")
+
+  script(type="text/javascript").
+    var DEFAULT_THEME = '#{DEFAULT_THEME}';
```
calzoneman, line 15: Didn't you mention in your Pug upgrade PR that interpolation is no longer supported?  Or is it still for JS string literals?

calzoneman, line 15: I understand that you tested it, I just want to understand what's going on because I'm a little confused.

Xaekai:
Interpolation is only no longer supported inside **element attributes**.

calzoneman:
Oh good, I love non-orthogonal language features

PR #596, opened 2016-07-11
Move ACP nav entry to templating

### LINKED ISSUE #516: ACP link is missing on home page
The ACP link in the navbar should either always or never be there depending on the user rank, the server should be adding it, not code running on the client.

Xaekai:
Closes #516

### DIFF
--- a/templates/nav.pug
+++ b/templates/nav.pug
@@ -36,6 +36,13 @@ mixin navdefaultlinks(page)
         li: a(href=loginDomain+"/login?dest=" + encodeURIComponent(baseUrl + page)) Login
         li: a(href=loginDomain+"/register") Register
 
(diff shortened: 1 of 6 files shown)


templates/nav.pug
```diff
@@ -36,6 +36,13 @@ mixin navdefaultlinks(page)
         li: a(href=loginDomain+"/login?dest=" + encodeURIComponent(baseUrl + page)) Login
         li: a(href=loginDomain+"/register") Register
 
+mixin navsuperadmin(page)
+  if superadmin
+    if page == '/'
```
calzoneman: I think it'd make more sense for this to be a boolean `newTab` or something, since you're not actually doing anything with the page path.

Xaekai:
Updated.

PR #598, opened 2016-07-12
Image strip option

Xaekai:
Allow users to strip images from chat to allay bandwidth issues and fears of IP capture.

### DIFF
--- a/templates/useroptions.pug
+++ b/templates/useroptions.pug
@@ -113,6 +113,7 @@ mixin us-chat
             option(value="always") Always
       +rcheckbox("us-sendbtn", "Add a send button to chat")
       +rcheckbox("us-no-emotes", "Disable chat emotes")
+      +rcheckbox("us-strip-image", "Remove images from chat")
 
 mixin us-mod
   #us-mod.tab-pane
--- a/www/js/data.js
+++ b/www/js/data.js
@@ -67,6 +67,7 @@ var FILTER_TO = 0;
 var NO_STORAGE = typeof localStorage == "undefined" || localStorage === null;
 var SOCKETIO_CONNECT_ERROR_COUNT = 0;
 var HAS_CONNECTED_BEFORE = false;
+var IMAGE_MATCH = /<img\s[^>]*?src\s*=\s*['\"]([^'\"]*?)['\"][^>]*?>/gi;
 
 function getOpt(k) {
     var v = NO_STORAGE ? readCookie(k) : localStorage.getItem(k);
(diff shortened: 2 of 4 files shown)


calzoneman:
> strap :100:

www/js/data.js
```diff
@@ -67,6 +67,7 @@ var FILTER_TO = 0;
 var NO_STORAGE = typeof localStorage == "undefined" || localStorage === null;
 var SOCKETIO_CONNECT_ERROR_COUNT = 0;
 var HAS_CONNECTED_BEFORE = false;
+var IMAGE_MATCH = /<img\s[^>]*?src\s*=\s*['\"]([^'\"]*?)['\"][^>]*?>/gi;
```
calzoneman, line 70: Browsers also permit using `src=`whatever`` (grave accent quotes), as well as `src=whatever` (no quotes).

www/js/util.js
```diff
@@ -1433,6 +1435,13 @@ function sendVideoUpdate() {
 
 /* chat */
 
+function stripImages(msg){
+    if (!USEROPTS.strip_image) {
+        return msg;
+    }
+    return msg.indexOf('<img') > -1 ? msg.replace(IMAGE_MATCH, "$1") : msg;
```
calzoneman: I don't think the `indexOf` check optimizes anything here, both `indexOf` and the regex match are effectively going to do a linear scan until `<img` is located, since there is no branching in the regex up to that point.

calzoneman: Also it might be useful to replace the image with a clickable link instead of just URL text.

Xaekai: it should be noted that the XSS seems to automatically lowercase all tags in filters, so trying to use `` to get around this isn't of concern.

Edit: Alright, if you think the optimization isn't, I'll remove it. I'm just trying to make sure, since this is part of the most commonly called function in the entire site

Xaekai: I think that would double up the linkage for most channels, since this only captures the img, and the img is usually already wrapped in an anchor in most channels img embed codes. Also, doesn't the built in link filter happen last anyway? That would effectively bypass the channel setting in the admin tab

www/js/data.js
Xaekai, line 70: Our XSS doesn't. Grave quoted and bare strings both get wrapped in double quotes.

Xaekai:
Updated. The replace function now wraps in an anchor if the channel setting for links is active.

templates/useroptions.pug
```diff
@@ -113,6 +113,7 @@ mixin us-chat
             option(value="always") Always
       +rcheckbox("us-sendbtn", "Add a send button to chat")
       +rcheckbox("us-no-emotes", "Disable chat emotes")
+      +rcheckbox("us-strip-image", "Strip external images")
```
calzoneman: Can this be renamed to like "Remove images from chat messages" or something?  "Strip external images" is pretty technical for the average CyTube user.  Other than that, LGTM based on local testing.

calzoneman:
I will certainly be making use of this feature :^)

Xaekai:
Done.

PR #599, opened 2016-07-12
Greatly improve performance of channel settings emote list

### LINKED ISSUE #594: Performance issues with large number of emotes
Creating this to track the work needed for this.  One channel has over 2,500 emotes, which causes it to lag a lot on channel load and occasionally trigger Firefox's "unresponsive script" dialog.

calzoneman:
The channel settings emote list is now paginated and leverages the same
basic code as the emote browser, but with a different renderer.  Fixes
 #594 and kills an ugly function.

### DIFF
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -2477,58 +2477,6 @@ function formatCSChatFilterList() {
     });
 }
 
-function formatCSEmoteList() {
-    var tbl = $("#cs-emotes table");
-    tbl.find("tbody").remove();
-    var entries = CHANNEL.emotes || [];
-    entries.forEach(function (f) {
-        var tr = $("<tr/>").appendTo(tbl);
-        var del = $("<button/>").addClass("btn btn-xs btn-danger")
-            .appendTo($("<td/>").appendTo(tr));
-        $("<span/>").addClass("glyphicon glyphicon-trash").appendTo(del);
-        del.click(function () {
-            socket.emit("removeEmote", f);
-        });
-        var name = $("<code/>").text(f.name).addClass("linewrap")
-            .appendTo($("<td/>").appendTo(tr));
-        var image = $("<code/>").text(f.image).addClass("linewrap")
-            .appendTo($("<td/>").appendTo(tr));
-        image.popover({
-            html: true,
-            trigger: "hover",
-            content: '<img src="' + f.image + '" class="channel-emote">'
-        });
-
-        image.click(function () {
-            var td = image.parent();
-            td.find(".popover").remove();
-            image.detach();
-            var edit = $("<input/>").addClass("form-control").attr("type", "text")
-                .appendTo(td);
-
-            edit.val(f.image);
-            edit.focus();
-
-            var finish = function () {
-                var val = edit.val();
-                edit.remove();
-                image.appendTo(td);
-                socket.emit("updateEmote", {
-                    name: f.name,
-                    image: val
-                });
-            };
-
-            edit.blur(finish);
-            edit.keydown(function (ev) {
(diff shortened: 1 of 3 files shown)


www/js/util.js
```diff
@@ -3112,3 +3060,95 @@ function stopQueueSpinner(data) {
...
+CSEmoteList.prototype.loadPage = function (page) {
+    var tbody = this.table.children[1];
+    tbody.innerHTML = "";
+
+    var start = page * this.itemsPerPage;
+    if (start >= this.emotes.length) {
+        return;
+    }
+    var end = Math.min(start + this.itemsPerPage, this.emotes.length);
+    var self = this;
```
Xaekai, line 3032: Closing the edit dialogue jumps you back to page 1. :rage: That's gonna piss me off (and probably others) to no end. It's especially annoying since it happens even if you haven't modified it.

`var self = this; this.page = page;`

www/js/util.js
```diff
@@ -3112,3 +3060,95 @@ function stopQueueSpinner(data) {
...
+
+    for (var i = start; i < end; i++) {
+        var row = document.createElement("tr");
+        tbody.appendChild(row);
+
+        (function (emote) {
+            // Add delete button
+            var tdDelete = document.createElement("td");
+            var btnDelete = document.createElement("button");
```
Xaekai, line 3042: `btnDelete` has no click handler to actually delete an emote. :smirk: I do think when you add that, Henke's suggestion was good: Populate the new emote form with the values of the deleted emote in case of accidental click, since you can't stack Bootstrap modals without hacks or additional libraries to make a confirmation dialog.

Xaekai:
Damn, you beat me to it. Of course your code looks much nicer than what I would have thought of initially :disappointed_relieved: 

Longstanding annoyance (affects the emotelist too): The prev `<<` and next `>>` buttons of `NewPaginator` should be their own blocks with `.float-left` and `.float-right` respectively, so they are in a _consistent_ position so you can spam click them to get where you want to be. Currently the width of the button block changes a lot because of the changing button sizes (especially from page 1 to page 13), and whether break buttons `...`  appear.. I was planning on making a PR for this myself, but thought I'd mention it.

www/js/util.js
```diff
@@ -3112,3 +3060,95 @@ function stopQueueSpinner(data) {
...
+            btnDelete.className = "btn btn-xs btn-danger";
+            var pennJillette = document.createElement("span");
+            pennJillette.className = "glyphicon glyphicon-trash";
+            btnDelete.appendChild(pennJillette);
+            tdDelete.appendChild(btnDelete);
+            row.appendChild(tdDelete);
+
+            // Add emote name
+            // TODO: editable
```
Xaekai, line 3057: This seems like busy work I could handle after this is merged.

Xaekai:
While you're at it, you could move showChannelSettings, startQueueSpinner, and stopQueueSpinner (lines 3017-3062) right above EmoteList, that way the emote list classes are directly adjacent to each other and it doesn't hurt the organization of the rest of the file any. Just an idea.

www/js/util.js
calzoneman, line 3042: This is what happens when I get distracted by network outages and forget what I was doing.

calzoneman:
The client is going to be factored out to use webpack or something similar to break it down into smaller modules.  At that time, EmoteList and CSEmoteList will become separate files in the source tree that depend one on the other.

calzoneman:
Pushing to beta, the paginator button spacing issue can be the subject of a different change.

Xaekai:
Tested. Looks ready for merge.

PR #603, opened 2016-07-16
Add a service socket

Xaekai:
to enable out of band access to the process commandline

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,6 +45,7 @@ process.stdin.on("data", function (data) {
     }
 });
 
+var net = require('net');
 function handleLine(line) {
     if (line === "/reload") {
         Logger.syslog.log("Reloading config");
@@ -75,5 +76,34 @@ function handleLine(line) {
         }
     } else if (line.indexOf("/reload-partitions") === 0) {
         sv.reloadPartitionMap();
+    } else if (line.indexOf("/globalban") === 0) {
+        var args = line.split(/\s+/); args.shift();
+        if (args.length >= 2 && net.isIP(args[0]) !== 0) {
+            var ip = args.shift();
+            var comment = args.join(' ');
+            require("./lib/database").globalBanIP(ip, comment, function (err, res) {
+                if (!err) {
+                    Logger.eventlog.log("[acp] " + "SYSTEM" + " global banned " + ip);
+                }
+            })
+        }
+    } else if (line.indexOf("/unglobalban") === 0) {
+        var args = line.split(/\s+/); args.shift();
+        if (args.length === 1 && net.isIP(args[0]) !== 0) {
+            var ip = args.shift();
+            require("./lib/database").globalUnbanIP(ip, function (err, res) {
+                if (!err) {
+                    Logger.eventlog.log("[acp] " + "SYSTEM" + " un-global banned " + ip);
+                }
+            })
+        }
     }
+}
+
+// Go Go Gadget Service Socket
+if (Config.get("service-socket.enabled")) {
+    Logger.syslog.log("Opening service socket");
+    var ServiceSocket = require('./lib/servsock');
+    var server = new ServiceSocket;
+    server.init(handleLine, Config.get("service-socket.socket"));
 }
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ process.stdin.on("data", function (data) {
     }
 });
 
-var net = require('net');
+var validIP = require('net').isIP;
 function handleLine(line) {
     if (line === "/reload") {
         Logger.syslog.log("Reloading config");
@@ -78,7 +78,7 @@ function handleLine(line) {
         sv.reloadPartitionMap();
     } else if (line.indexOf("/globalban") === 0) {
         var args = line.split(/\s+/); args.shift();
-        if (args.length >= 2 && net.isIP(args[0]) !== 0) {
+        if (args.length >= 2 && validIP(args[0]) !== 0) {
             var ip = args.shift();
             var comment = args.join(' ');
             require("./lib/database").globalBanIP(ip, comment, function (err, res) {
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -75,5 +75,33 @@ function handleLine(line) {
...
+        // simple validation for IPv4
+        // TODO: IPv6
+        if(args.length === 2 && args[0].match(/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/)){
+            require("./lib/database").globalBanIP(args[0], args[1], function (err, res) {
+                if (!err) {
+                    Logger.eventlog.log("[acp] " + "SYSTEM" + " global banned " + args[0]);
+                }
+            })
+        }
+    } else if (line.indexOf("/clearban") === 0) {
```
calzoneman: /unglobalban or something like that would be more consistent naming.

Replace the IP address regex with `net.isIPv4()`: https://nodejs.org/api/net.html#net_net_isipv4_input

index.js
```diff
@@ -75,5 +75,33 @@ function handleLine(line) {
...
+                if (!err) {
+                    Logger.eventlog.log("[acp] " + "SYSTEM" + " un-global banned " + args[0]);
+                }
+            })
+        }
     }
 }
+
+// Go Go Gadget Service Socket
+if(Config.get("service-socket.enabled")){
```
calzoneman: Spacing around  `if () {`

index.js
```diff
@@ -75,5 +75,33 @@ function handleLine(line) {
...
+    Logger.syslog.log("Opening service socket");
+    var ServiceSocket = require('./lib/servsock');
+    var server = new ServiceSocket;
```
calzoneman, line 119: Is this even legal in ECMAScript?

servcmd.sh.js
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env node
+/*
+**  CyTube Service Socket Commandline
+*/
+
+var Config = require("./lib/config");
+Config.load("config.yaml");
+
+if(!Config.get("service-socket.enabled")){
+	console.error('The Service Socket is not enabled.');
```
calzoneman: INDENTING WITH TABS REEEEEEEEEEEEEEEE

servcmd.sh.js
```diff
@@ -0,0 +1,56 @@
...
+
+        if(msg === '__disconnect'){
+            console.log('Server shutting down. Terminating.')
+            return cleanup();
+        }
+
+        // Generic message handler
+        console.info('Server:', data)
+    })
+    .on('error', (data)=>{
```
calzoneman: Spacing around arrow operator

src/servsock.js
```diff
@@ -0,0 +1,54 @@
...
+                }
+                return this.openServiceSocket();
+            });
+        });
+    }
+
+    openServiceSocket(){
+        this.server = net.createServer((stream)=>{
+            let self = Date.now();
+            this.connections[self] = (stream);
```
calzoneman: Why is `stream` wrapped in parens

src/servsock.js
```diff
@@ -0,0 +1,54 @@
...
+                    console.error(err); process.exit(0);
+                }
+                return this.openServiceSocket();
+            });
+        });
+    }
+
+    openServiceSocket(){
+        this.server = net.createServer((stream)=>{
+            let self = Date.now();
```
calzoneman: Throughout CyTube, `self` is traditionally used as a lexical closure around `this` for legacy ECMAScript 5 code where JavaScript would mangle `this` in callbacks/nested functions.  Using it as something unrelated here seems like a weird naming choice.

src/servsock.js
```diff
@@ -0,0 +1,54 @@
...
+            this.connections[self] = (stream);
+            stream.on('end', ()=>{
+                delete this.connections[self];
+            });
+            stream.on('data', (msg)=>{
+                this.handler(msg.toString());
+            });
+        })
+        .listen(this.socket);
```
calzoneman: Convention throughout CyTube is to have chained calls start on the same line

src/servsock.js
Xaekai: Vestigial.

Xaekai:
Updated.

index.js
```diff
@@ -75,5 +76,31 @@ function handleLine(line) {
         }
     } else if (line.indexOf("/reload-partitions") === 0) {
         sv.reloadPartitionMap();
+    } else if (line.indexOf("/globalban") === 0) {
+        var args = line.split(" "); args.shift();
+        if (args.length === 2 && net.isIPv4(args[0]) || net.isIPv6(args[0])) {
```
calzoneman: Operator precedence means that this is parsed as `(args.length === 2 && net.isIPv4(args[0])) || net.isIPv6(args[0])`, also if you want to support both you can just check if `net.isIP()` is nonzero.

Xaekai:
Updated and retested.

index.js
```diff
@@ -45,6 +45,7 @@ process.stdin.on("data", function (data) {
     }
 });
 
+var net = require('net');
```
calzoneman: Remove unused import

src/servsock.js
```diff
@@ -0,0 +1,53 @@
+var fs = require('fs');
+var net = require('net');
+
+export default class ServiceSocket {
```
calzoneman, line 4: Does this class need error event handlers?  In case the client connection breaks or something.

index.js
Xaekai: It's not unused. It's used to to validate user input, which was your suggestion over a regex.

Edit: I can change it to just import the one function though.

src/servsock.js
Xaekai, line 4: I have thus far been unable to break anything. Sending SIGKILL to the client leaves orphaned children (the tail on the log files) but doesn't affect the service at all. Sending SIGTERM to the service after SIGKILL'ing a client shuts down the service with no errors.

PR #609, opened 2016-08-04
Use explicit GitHub URL in package.json

jarrpa:
The dependency on status-message-polyfill appeared to be missing part of its URL, judging by the rest of the file. Changing this line allowed me to build and install on a self-hosted server.


calzoneman:
I believe this issue is also caused by you using an older version of npm/node.js that does not support this syntax for github dependencies.  I'm ok with merging this because it's a pretty harmless change that brings consistency with the other github dependencies in package.json, but as noted on your other PR, I do not support old versions of node.js/npm.

PR #610, opened 2016-08-04
Send the right account object on user login

jarrpa:
Don't send a null "account" object, instead use user.account. Was getting an error about this when trying to log in to my admin account on a self-hosted installation.


calzoneman:
The `account` object passed in the callback will not be null unless the `err` is non-null, in which case the code already handles this correctly.

The reason you are seeing a null `account` is because you are using an older version of node.js which does not pass arguments to callbacks from `process.nextTick(cb, arg1, arg2, ...)`.

Per the [wiki](https://github.com/calzoneman/sync/wiki/CyTube-3.0-Installation-Guide#nodejs), node.js v4.x LTS and stable (currently v6.x) are supported.  I no longer support v0.10.x and v0.12.x, which are older versions of node.js.

PR #614, opened 2016-08-11
Restrict chat messages from newer accounts/IPs

calzoneman:
- Channel owners can now require accounts to have a certain "age" to be able to chat, or post links in chat
  - Default: 10 minutes before chatting, 1 hour before posting images
- Age is determined by the user's account registration time (if logged in), and the time specified in their "IP session" cookie which is set when they first visit any channel, whichever is older.


calzoneman:
Now live on beta.synchtube.me

Xaekai:
Dank

PR #633, opened 2016-10-24
Added Feature: Desktop Notifications

dinsfire64:
Added desktop notification for chat. Allows for notifications when tab/window is not in focus on mention of username or all messages. Allows for mobile users to keep cytube open in the background and receive notifications of mentions or all messages.


www/js/ui.js
```diff
@@ -1,11 +1,12 @@
-/* window focus/blur */
-$(window).focus(function() {
-    FOCUSED = true;
-    clearInterval(TITLE_BLINK);
-    TITLE_BLINK = false;
-    document.title = PAGETITLE;
-}).blur(function() {
-    FOCUSED = false;
+document.addEventListener("visibilitychange", function() {
```
calzoneman, line 1: What was the motivation for this change?  I can think of at least a couple issues that need clarification:
- The [spec](https://www.w3.org/TR/page-visibility/) for `visibilityState` is relatively recent (2013).  There are a number of users who will have browsers too old to support this (people using old Androids / built in browsers on game consoles / etc.).  It would be better to at least fall back to the old `focus` event for these browsers.
- It seems like according to the spec, this behaves differently than the code that was there before.
  
  > For example, in the following cases the visibilityState attribute would return the DOMString hidden:
  > 
  > The User Agent is minimized.
  > The User Agent is not minimized, but the page is on a background tab.
  > The Operating System lock screen is shown.
  
  How does this behave when the window is fully visible, but not selected?  For example, if I have a browser on one monitor, and a game on the other, and the game is focused?

www/js/util.js
```diff
@@ -688,10 +690,42 @@ function saveUserOptions() {
...
 
+    getNotificationPermission();
+
     storeOpts();
     applyOpts();
 }
 
+function getNotificationPermission() {
+    //Only ask for permission if they desire the permission.
+    if(USEROPTS.desktop_notification != "never")
```
calzoneman, line 701: Braces on the same line

``` js
if (...) {
```

Use `!==` rather than `!=` for consistency (even though it won't cause any type coercion issues in this specific scenario)

www/js/util.js
```diff
@@ -688,10 +690,42 @@ function saveUserOptions() {
...
+    {
+        notificationsAvailable = false;
```
calzoneman, line 703:
``` js
var notificationsAvailable = false;
```

www/js/util.js
```diff
@@ -688,10 +690,42 @@ function saveUserOptions() {
...
+        
+        //Check to see if notifications are available.
+        if (!Notification) 
+        {
+            alert('Desktop notifications are not availble on your system, please update your browser.');
```
calzoneman, line 708: available*

www/js/util.js
```diff
@@ -688,10 +690,42 @@ function saveUserOptions() {
...
+            alert('Desktop notifications are not availble on your system, please update your browser.'); 
+            USEROPTS.desktop_notification = "never";
+        } else {
+            //Ask for permission if we don't have it.
+            if (Notification.permission !== "granted")
+            {
+                Notification.requestPermission();
+            }
+            
+            notificationsAvailable = Notification.permission == "granted";
```
calzoneman, line 717: `===`

www/js/util.js
```diff
@@ -1625,6 +1659,12 @@ function addChatMessage(data) {
     }
 
     pingMessage(isHighlight);
+    
+    //Strip HTML before passing it to the notification
```
calzoneman, line 1663: This belongs in `showMessage()` since the need for it is local to that.  Also allows you to optimize and not create an unnecessary DOM element for the conversion unless notifications are enabled and this message would result in one.

www/js/util.js
```diff
@@ -1642,7 +1682,53 @@ function trimChatBuffer() {
...
+            if(USEROPTS.desktop_notification == "always" || 
+               USEROPTS.desktop_notification == "onlyping" && nameMentioned) {
+                   
+               notificationTitle = "/r/" + CHANNEL.name;
+               iconLocation = "/favicon.ico";
+               
+               //There are two ways to send notifications,
+               //service worker only works over HTTPS, so let's try that first.
+               try {
+                    navigator.serviceWorker.register('/sw.js');
```
calzoneman, line 1705: Does this need to be done on every message or only once?  Also, consistency of single vs. double quotes.

www/js/util.js
```diff
@@ -1642,7 +1682,53 @@ function trimChatBuffer() {
...
+                
+                    navigator.serviceWorker.ready.then(function(registration) {
+                        registration.showNotification(notificationTitle, {
+                            body: message,
+                            icon: iconLocation,
+                        });
+                    });
+                } catch(e) {
```
calzoneman, line 1713: Log the error.

``` js
console.error('Could not send notification with serviceWorker:', e.stack);
```

www/js/util.js
```diff
@@ -1642,7 +1682,53 @@ function trimChatBuffer() {
...
+                    
+                    //If we can't use service worker then we need to default to the old way.
+                    //This will be depricated in the future
+                    
+                    var notificationOptions = {
+                        body: notificationTitle,
```
calzoneman, line 1719: I think there is a bug here where the notification only shows the title and not the actual message.

www/js/util.js
```diff
@@ -1642,7 +1682,53 @@ function trimChatBuffer() {
...
+                    navigator.serviceWorker.ready.then(function(registration) {
+                        registration.showNotification(notificationTitle, {
+                            body: message,
+                            icon: iconLocation,
+                        });
+                    });
+                } catch(e) {
+                    
+                    //If we can't use service worker then we need to default to the old way.
+                    //This will be depricated in the future
```
calzoneman, line 1716: Deprecated*

calzoneman (changes requested):
- What is this empty `sw.js` for?
- Please [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits down to one to avoid clutter.

www/js/util.js
calzoneman: The reason being I guarantee at some point someone will have a problem and come to IRC, and it's easiest to debug when you can ask them for a screenshot of the JS console and immediately know where something failed.

calzoneman:
Closing due to inactivity

PR #636, opened 2016-12-14
HTTPS enforcement phase 1

calzoneman:
Phase 1 of HTTPS enforcement:

  * All new external CSS links must be HTTPS
  * All new external JS links must be HTTPS
  * All new profile image links must be HTTPS
  * Existing HTTP links remain untouched

This needs some testing in beta, haven't deployed it yet.

During Phase 2, existing HTTP links will be phased out by prompting users to change them (with a warning for the cutoff date).  Phase 3 actually enforces HTTPS by invalidating all non-HTTPS links for these 3 components.

Warning/enforcement of raw video links will be handled in a different commit.  Emotes/inline images will be handled separately (I haven't decided how yet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "Calvin Montgomery",
   "name": "CyTube",
   "description": "Online media synchronizer and chat",
-  "version": "3.24.3",
+  "version": "3.25.0",
   "repository": {
     "url": "http://github.com/calzoneman/sync"
   },
--- a/src/channel/opts.js
+++ b/src/channel/opts.js
@@ -176,8 +176,11 @@ OptionsModule.prototype.handleSetOptions = function (user, data) {
 
         var link = data.externalcss.substring(0, 255).trim();
         if (!link) {
+            sendUpdate = (this.opts.externalcss !== "");
             this.opts.externalcss = "";
-            sendUpdate = true;
+            user.socket.emit("validationPassed", {
+                target: "#cs-externalcss"
+            });
         } else {
             var data = url.parse(link);
             if (!data.protocol || data.protocol !== 'https:') {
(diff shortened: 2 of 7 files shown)


Xaekai:
I think enforcing https on the channels dynamic inline content would be impractical. Everything else seems perfectly reasonable.

calzoneman:
Change is live on beta.

irufus:
I saw this in action on v4c. Any reason why you're not force upgrading the client on access?

calzoneman:
>I saw this in action on v4c

Sorry for the confusion, I meant to write up some documentation/FAQ but the announcement went up prematurely.

>Any reason why you're not force upgrading the client on access?

This change is referring specifically to *external* links to CSS/JS.  Inline CSS/JS will continue to be served via the CyTube socket, and will not require any changes from channel owners.

The end goal is for all access to CyTube to be enforced over HTTPS via the `Strict-Transport-Security` header, which directs users' browsers to use https://cytu.be.  However, browser security policy prohibits loading active content (such as CSS/JS) from a plain HTTP URL when the page itself is HTTPS.  Hence, channels' external content will need to be migrated first, otherwise it will stop working once CyTube goes mandatory HTTPS.

The information/FAQ I post later might answer any more questions you have.

calzoneman:
http://blog.cytu.be/2016-12-19-https.html

SysadminJeroen:
One thing that I'm recommending to enforce HTTPS is when someone includes an image from a site that does not support HTTPS as an emoticon, CyTube should use https://github.com/atmos/camo on a subdomain and pass the images through the proxy. The proxy will anonymize the traffic and it'll prevent mixed content warnings. GitHub is using Camo for mixed content as well and I have been using it on a phpBB installation and it just works.

calzoneman:
That is an interesting idea that I'll have to look into, so thanks for sharing.  I've always liked the idea of proxying user avatars and chat emotes, but I've always been concerned about (1) potential bandwidth costs, and (2) people abusing it to host illegal content.

SysadminJeroen:
All it does is proxy the content through a webserver that actually does use HTTPS. So no illegal content emails due to no content actually getting hosted on the server. I mean they don't send an email to all the Tier 1 providers just because they transferred the illegal content from one provider through their network to another provider.


And the amount of bandwidth depends.
Some browsers cache images.
I don't really have anything else to say, as I do not know how much bandwidth CyTube will use nor how much bandwith CyTube has.

PR #648, opened 2017-03-06
Change fallback voteskip ratio to .5 from 0

deerfarce:
moderators might make a mistake changing the skip ratio, causing it to fall back to 0% and allowing users to "skiptrain" before it is fixed

### DIFF
--- a/src/channel/opts.js
+++ b/src/channel/opts.js
@@ -116,7 +116,7 @@ OptionsModule.prototype.handleSetOptions = function (user, data) {
         if (isNaN(ratio) || ratio < 0) {
             user.socket.emit("validationError", {
                 target: "#cs-voteskip_ratio",
-                message: "Input must be a number 0 or greater. 1.0 is 100%."
+                message: `Input must be a number 0 or greater, not "${data.voteskip_ratio}"`
             });
         } else {
             this.opts.voteskip_ratio = ratio;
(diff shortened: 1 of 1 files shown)


calzoneman:
I think actually there shouldn't be a fallback &mdash; if the user tries to send invalid input, they should get an error response.

[Here](https://github.com/calzoneman/sync/blob/3.0/src/channel/opts.js#L168-L175)'s an example of how it's done for CSS (although I just noticed the if-block there is missing an early exit, so I'll have to fix that).

calzoneman:
Are you planning to update this PR based on my above comment, or would you like me to close it and treat it as a bug report that needs to be fixed separately?

deerfarce:
Oh, sorry, I forgot about this. I will update the PR this evening when I can.

deerfarce:
Updated.

src/channel/opts.js
```diff
@@ -114,10 +114,17 @@ OptionsModule.prototype.handleSetOptions = function (user, data) {
     if ("voteskip_ratio" in data) {
         var ratio = parseFloat(data.voteskip_ratio);
         if (isNaN(ratio) || ratio < 0) {
-            ratio = 0;
+            user.socket.emit("validationError", {
+                target: "#cs-voteskip_ratio",
+                message: "Input must be a number 0 or greater. 1.0 is 100%."
```
calzoneman: Just a nitpick, but I like to get in the habit of including the erroneous value just to make it more obvious when something is wrong.  For example:

```
message: `Input must be a number 0 or greater, not "${data.voteskip_ratio}"`
```

Other than that, it looks fine to me.

PR #652, opened 2017-03-21
Add upper limit to poll options

h2v4c:
Found a bug that caused server issues if a poll with 100k options was made.

### DIFF
--- a/src/poll.js
+++ b/src/poll.js
@@ -6,6 +6,9 @@ var Poll = function(initiator, title, options, obscured) {
     title = XSS.sanitizeText(title);
     this.title = title.replace(link, "<a href=\"$1\" target=\"_blank\">$1</a>");
     this.options = options;
+    if(this.options.length>100) {
+        return;
+    }
     for (var i = 0; i < this.options.length; i++) {
         this.options[i] = XSS.sanitizeText(this.options[i]);
         this.options[i] = this.options[i].replace(link, "<a href=\"$1\" target=\"_blank\">$1</a>");


calzoneman:
I appreciate the PR, but in the future, please contact me privately about this kind of issue since this is a Denial of Service vector and I'd like to keep the audience to a minimum until the patch is deployed (refer to [this document](https://github.com/CyTube/cytu.be-policies/blob/master/security.md#reporting-a-security-issue) for details).  I'm also not sure that this PR is a sufficient fix to the issue, because it early-exits from a constructor without throwing any exception, so it leaves the poll in an inconsistent state that may trigger other issues due to violating the contract (also, `this.options` is already assigned when this condition is reached).

I'm probably going to go a little more in-depth on this one, including adding proper clientside limits and errors as well to make sure no one accidentally breaching this limit is confused.  The error handling is not great in the poll module, and I'll need to audit the other places user input is accepted for any other places 4-years-ago-me didn't bother adding validation based on resource limits (in this case, the number of poll items didn't directly crash the process, but caused it to allocate too much memory and get killed by the kernel).

calzoneman:
I'll be working on a fix after dinner, but if there are any server admins reading this who are impacted, a short term mitigation would be to disable the poll module by commenting out [this line](https://github.com/calzoneman/sync/blob/3.0/src/channel/channel.js#L155).

calzoneman:
Closing via https://github.com/calzoneman/sync/commit/9dc82ad444c14b065b29393ac9d056805fc7ac27

PR #664, opened 2017-05-16
Emote renaming

### LINKED ISSUE #557: Ability to rename emotes
Hello,

I was wondering if in the future there could possibly be an option to rename emotes,
similar to how you can modify chat filters.

Additionally, any plans for categorizing emotes in the future?

Thanks!

Xaekai:
This allow emotes to be renamed in the same fashion the image URLs can be changed.

Closes #557

### DIFF
--- a/src/channel/emotes.js
+++ b/src/channel/emotes.js
@@ -168,12 +168,11 @@ EmoteModule.prototype.handleRenameEmote = function (user, data) {
         return;
     }
 
+    this.emotes.renameEmote(Object.assign({}, f));
+
     var chan = this.channel;
     chan.broadcastAll("renameEmote", f);
     chan.logger.log(`[mod] ${user.getName()} renamed emote: ${f.old} -> ${f.name}`);
-
-    // Doing this after the broadcast because this function deletes the "old" prop
-    this.emotes.renameEmote(f);
 };
 
 EmoteModule.prototype.handleUpdateEmote = function (user, data) {
--- a/src/channel/emotes.js
+++ b/src/channel/emotes.js
@@ -20,14 +20,28 @@ EmoteList.prototype = {
         this.emotes = Array.prototype.slice.call(emotes);
     },
 
+    emoteExists: function (emote){
+        if(this.emotes.filter((item)=>{ return item.name === emote.name }).length){
+            return true;
+        }
+        return false;
+    },
+
     renameEmote: function (emote) {
+        var found = false;
         for (var i = 0; i < this.emotes.length; i++) {
             if (this.emotes[i].name === emote.old) {
+                found = true;
                 this.emotes[i] = emote;
                 delete this.emotes[i].old;
                 break;
             }
         }
+
+        if(found){
+            return true;
+        }
+        return false;
     },
 
     updateEmote: function (emote) {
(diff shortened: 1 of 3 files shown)


src/channel/emotes.js
```diff
@@ -133,6 +144,38 @@ EmoteModule.prototype.sendEmotes = function (users) {
...
+            alert: true
+        });
+        return;
+    }
+
+    var chan = this.channel;
+    chan.broadcastAll("renameEmote", f);
+    chan.logger.log(`[mod] ${user.getName()} renamed emote: ${f.old} -> ${f.name}`);
+
+    // Doing this after the broadcast because this function deletes the "old" prop
```
calzoneman: Prefer explicitly cloning the object instead of mutating it.  Immutable objects are much easier to reason about.

www/js/callbacks.js
```diff
@@ -1019,6 +1019,54 @@ Callbacks = {
...
+        var badBefore = /\s/g.test(data.old);
+        var badAfter = /\s/g.test(data.name);
+        var oldName = data.old;
+        delete data.old;
+
+        data.regex = new RegExp(data.source, "gi");
+
+        for (var i = 0; i < CHANNEL.emotes.length; i++) {
+            if (CHANNEL.emotes[i].name === oldName) {
+                CHANNEL.emotes[i] = data;
```
calzoneman, line 1032: Doesn't the non-bad emote map also need to be updated?

www/js/callbacks.js
```diff
@@ -1019,6 +1019,54 @@ Callbacks = {
...
+                break;
+            }
+        }
+
+        // Now bad
+        if(badAfter){
+            // But wasn't bad before: Add it to bad list
```
calzoneman, line 1039: I've been meaning to separate out all the emote logic into a shared interface that can be implemented by the old method, and the new method, and then phase the old one out.  That would simplify this a lot.

www/js/util.js
```diff
@@ -3120,13 +3120,67 @@ CSEmoteList.prototype.loadPage = function (page) {
...
+                function save() {
+                    var val = editInput.value;
+                    tdName.removeChild(editInput);
+                    tdName.appendChild(nameDisplay);
+
+                    // Nothing was changed
+                    if(val === emote.name){ return }
+
+                    // Emote name already exists
+                    if( CHANNEL.emotes.filter((emote)=>{ return emote.name === val }).length ){
```
calzoneman: I'm not supporting ES6 in core browser side yet.

www/js/util.js
```diff
@@ -3120,13 +3120,67 @@ CSEmoteList.prototype.loadPage = function (page) {
...
             // Add emote name
-            // TODO: editable
             var tdName = document.createElement("td");
             var nameDisplay = document.createElement("code");
             nameDisplay.textContent = emote.name;
             tdName.appendChild(nameDisplay);
             row.appendChild(tdName);
 
+            var $nameDisplay = $(nameDisplay);
+            $nameDisplay.click(function (clickEvent) {
```
calzoneman, line 3130: This whole thing is kind of gross and reminds me that I want to redo the JS UI rendering

www/js/callbacks.js
Xaekai: I'm sure you've realized by now that is indeed covered a bit lower into the function.

www/js/util.js
Xaekai: I should have realized that. I will make it classic function.

www/js/callbacks.js
calzoneman: Ctrl+F `emoteMap` doesn't show anything in your changes.

Xaekai: Oh, I misread.

Xaekai: Resolved.

Xaekai:
Anything else I need to change?

PR #665, opened 2017-05-19
Fix max duration input bug

deerfarce:
lazily cast rawMax to string before processing it.

### DIFF
--- a/src/channel/opts.js
+++ b/src/channel/opts.js
@@ -173,7 +173,10 @@ OptionsModule.prototype.handleSetOptions = function (user, data) {
     }
 
     if ("playlist_max_duration_per_user" in data) {
-        const rawMax = data.playlist_max_duration_per_user+"";
+        var rawMax = data.playlist_max_duration_per_user;
+        if (typeof rawMax === "number") {
+            rawMax = rawMax.toString();
+        }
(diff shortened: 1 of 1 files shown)


Xaekai:
Do not be lazy with code in the server core

:angry:

deerfarce:
Improved it I think, works just fine.

calzoneman:
What bug is being fixed here exactly?  The option should accept exactly either a string or a number, not both.  So if the client is already parsing the timestring and sending the raw number, then the server should accept a number only.

Xaekai:
There is actually a bug somewhere, but I haven't had a chance to look for it myself, and have no idea if this PR actually fixes anything
image

deerfarce:
In Channel Settings, putting something in such as 00:00:59 will cause an error with the message, "Input must be a time in the format HH:MM:SS, not 59." The server receives a number, not a string, so it just passes right into the first if block and emits the validationError.

calzoneman:
I could have sworn I tested this before checking it in, but oh well.  It looks like indeed the UI converts the HH:MM:SS timestamp into a number before passing it to the server.  So I guess the parsing logic can be removed and just accept a number.  Easy fix.

calzoneman:
Fixed via https://github.com/calzoneman/sync/commit/55b03d51d70dd44bd756b67c516ee6d0f8431977

Issue #666, opened 2017-05-23
[User-reported] Chat is not auto-rescrolled following window resize event


calzoneman:
When resizing the window, the height of the chat buffer may change, which causes the scroll position to change.  The scroll position is not automatically reset to the bottom of the buffer until the next chat message is sent.

Should be easily fixable by listening to the resize event on the `window` object and calling `scrollChat()`.

calzoneman:
Also, issue #666 😈

Xaekai:
Damn it, I wanted #666 for a PR.

calzoneman:
Rekt.

PR #667, opened 2017-05-24
[ACP] Allow searching users by email.

### LINKED ISSUE #530: Search by user email in acp
Allow searching by email for users

Xaekai:
Closes #530

### DIFF
--- a/src/acp.js
+++ b/src/acp.js
@@ -91,15 +91,10 @@ function handleGlobalBanDelete(user, data) {
 }
 
 function handleListUsers(user, data) {
-    var query = data.query;
-    if (typeof query !== "string") {
-        query = "";
-    }
-
+    var value = data.value;
     var field = data.field;
-    if (typeof field !== "string") {
-        field = "name";
-    }
+    value = (typeof value !== 'string') ? '' : value;
+    field = (typeof field !== 'string') ? 'name' : field;
 
     var fields = ["id", "name", "global_rank", "email", "ip", "time"];
 
@@ -110,7 +105,7 @@ function handleListUsers(user, data) {
         return;
     }
 
-    db.users.search(field, query, fields, function (err, users) {
+    db.users.search(field, value, fields, function (err, users) {
         if (err) {
             user.socket.emit("errMessage", {
                 msg: err
--- a/src/acp.js
+++ b/src/acp.js
@@ -91,14 +91,26 @@ function handleGlobalBanDelete(user, data) {
 }
 
 function handleListUsers(user, data) {
-    var name = data.name;
-    if (typeof name !== "string") {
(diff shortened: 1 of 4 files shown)


Xaekai:
This should have been #666. 
image

src/database/accounts.js
```diff
@@ -51,29 +51,16 @@ module.exports = {
     },
 
     /**
-     * Search for a user by name
+     * Search for a user by any field
      */
-    search: function (name, fields, callback) {
-        /* This bit allows it to accept varargs
```
Xaekai, line 57: I couldn't find this actually been used anywhere. And wouldn't it be better to just package the vars as an object for such a use case?

src/acp.js
```diff
@@ -91,14 +91,26 @@ function handleGlobalBanDelete(user, data) {
 }
 
 function handleListUsers(user, data) {
-    var name = data.name;
-    if (typeof name !== "string") {
-        name = "";
+    var query = data.query;
```
calzoneman: "query" might be a confusing name for this, because it's overloaded with the concept of an SQL "query".  I might call it "searchTerm" or something like that.

src/database/accounts.js
calzoneman: "Varags" is poorly worded, I think my reasoning here was to make the `fields` parameter optional and default to just selecting the name.  But I don't know if this is actually used.

src/acp.js
Xaekai: well I originally had it as `field`, but thought that was confusing as well. I'll change it to searchTerm if it pleases the crown.

calzoneman: If I'm interpreting your code correctly, the "field" is the column being searched, whereas what you're calling "query" here is the value to look for.  Is that right?

Xaekai: yeah, that's it.

Xaekai: would `value` work better?

calzoneman: That would be fine.

src/database/accounts.js
Xaekai: Was destructuring unavailable when you wrote that bit? I've been looking for good excuses to make use of it, but never seem to find them.
```js
function destructuredVarArgs({ name, fields, callback }){
  fields = typeof fields === 'undefined' ? ['name'] : fields;
}
```

calzoneman: The last time `acp.js` was touched meaningfully (i.e. not just stripping comments or moving the file path) was 2014, so yeah that predates ES2015.

But if nothing is actually using the optionality of the fields, I think it's fine to not preserve that behavior and just make it a required field.  A quick grep shows nothing is using it except the ACP.

PR #670, opened 2017-05-29
Support proxying chat images via camo

calzoneman:
Camo: https://github.com/atmos/camo.  This has a couple advantages over
just allowing images to be dumped as-is:

  - Prevents mixed-content warnings by allowing the server to proxy HTTP
    images to an HTTPS camo instance
  - Protects users' privacy by not exposing their browser directly to
    the image host
  - Allows the camo proxy to intercept and reject bad image sources
    (URLs that are not actually images, gigapixel-sized images likely to
    DoS users' browsers, etc.)

Whitelisting specific domains is supported for cases where the source is
known to be trustworthy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ www/cache
 google-drive-subtitles
 lib/
 integration-test-config.json
+conf/*.toml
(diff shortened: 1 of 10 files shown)


Xaekai:
Does this cover the user profile avatars?

calzoneman:
No, I'll integrate that separately.  It can use the same library functions defined here though.

Xaekai:
I presume the `config.template.yaml` change is pending another commit? :smirk:

calzoneman:
I'm condering `config.yaml` as end-of-life and avoiding adding new config to it (in favor of smaller, more easily understandable files in `conf/`, and storing application logic configuration in the database eventually).

I haven't added the example configuration though.

PR #674, opened 2017-06-05
Touch up validations on account pages

Xaekai:


Xaekai:
If you think it's necessary I can break out the JS into a separate file, but I saw little need of it with how short it is.

templates/account-channels.pug
```diff
@@ -55,9 +55,49 @@ html(lang="en")
...
+            var value = box.val();
+            var lastkey = Date.now();
+            box.data("lastkey", lastkey);
+
+            setTimeout(function () {
+                if (box.data("lastkey") !== lastkey || box.val() !== value) {
+                    return;
+                }
+                if(nameIsInvalid(value)){
+                    $('#validation').text(nameIsInvalid(value));
```
calzoneman: https://github.com/calzoneman/sync/blob/3.0/templates/register.pug has some JS already to update the UI in a nice way using bootstrap's `has-error` and such.  Maybe that can be adapted for here.

templates/account-channels.pug
```diff
@@ -55,9 +55,49 @@ html(lang="en")
...
+                button#register.btn.btn-primary.btn-block(type="submit") Register
 
     include footer
     +footer()
+    script( type='text/javascript').
+        function nameIsInvalid(id){
+            if(/\s/.test(id)){
+                return 'Channel URLs may not contain spaces';
+            }
+            if(id.length > 30){
```
calzoneman: Add `maxlength="30"` to the `input` element

templates/account-channels.pug
```diff
@@ -55,9 +55,49 @@ html(lang="en")
                 input(type="hidden", name="_csrf", value=csrfToken)
                 input(type="hidden", name="action", value="new_channel")
                 .form-group
-                  label.control-label(for="channelname") Channel Name
-                  input#channelname.form-control(type="text", name="name")
-                button.btn.btn-primary.btn-block(type="submit") Register
+                  label.control-label(for="channelname") Channel URL
+                  span#validation.pull-right
+                  .input-group
+                    span.input-group-addon #{baseUrl}/r/
```
calzoneman: Not sure how `baseUrl` is calculated, does that potentially contain redundant ports and such?  When I thought of doing this, I was thinking of adding a new local variable to the template that's set to `req.host`.

Xaekai:
I think that should do it.

PR #676, opened 2017-06-09
Fuck off Google

Xaekai:
This will prevent Google from using the contents of the `#messagebuffer` as snippet text in it's search index


calzoneman:
https://stackoverflow.com/questions/14314111/avoid-crawling-part-of-a-page-with-googleoff-and-googleon#19489890

PR #679, opened 2017-06-12
Changes to enable dockerization and reverse proxying

davidsiaw:
This is a set of changes to sync that I used to make dockerizing it possible and simplifying deployment.

It also has a few changes that fix some bugs that cropped up when reverse proxying https and wss requests to cytube (that is, allowing both ports 8080 and 1337 to be accessed through a server that only has port 80 and 443 accessible through a reverse proxy via separate domain names).

There is a change to the cytubefilters repository reference as well, this repository has a change that enables cytubefilters to be built on Alpine Linux (which is a small distro that I wanted to use as the base of my docker container).

The config template has been changed to enable one to simply run `cat config.yaml.template | envsubst` to generate a proper config file. This is far from complete but modifies some of the more important variables in the config file and hopefully will go some ways to improving the quality of life for people intending to deploy cytube.

In addition I have written a Dockerfile that will describe a build of cytube that *should* work out of the box and is self contained for testing, greatly shortening the time required to set up a test server.

I use this set of changes live on a private sync server I use with my friends.

I'm opening this pull request to open discussions about the changes and hopefully some of the changes will make it in.


Xaekai:
1.) Why didn't you do this in a feature branch? You'll find it's a pain in the ass to work in the `master` (in this case `3.0`) branch when you plan to PR upstream. Especially if you plan on contributing again.
2.) Google "Squash Commits"

package.json
```diff
@@ -25,7 +25,7 @@
     "csrf": "^3.0.0",
     "cytube-common": "git://github.com/CyTube/cytube-common",
     "cytube-mediaquery": "git://github.com/CyTube/mediaquery",
-    "cytubefilters": "git://github.com/calzoneman/cytubefilters#67c7c69a",
+    "cytubefilters": "git://github.com/davidsiaw/cytubefilters",
```
Xaekai, line 28: This probably isn't going to fly. If you need to change something there to make this work, make a PR there.

config.template.yaml
```diff
@@ -216,7 +217,7 @@ channel-blacklist: []
 # server-synched raw file playback.  This requires the following:
 #   * ffmpeg must be installed on the server
 ffmpeg:
-  enabled: false
+  enabled: true
```
Xaekai, line 220: This is disabled by default for a reason.

davidsiaw:
Because I didn't plan to PR upstream before

¯\_(ツ)_/¯

config.template.yaml
```diff
@@ -4,11 +4,11 @@
 # user: username to authenticate as
 # password: password for user
 mysql:
-  server: 'localhost'
-  port: 3306
-  database: 'cytube3'
-  user: 'cytube3'
-  password: ''
+  server: '${MYSQL_HOST}'
```
Xaekai, line 7: Is there any way you could make Docker use a separate template file?

All these Docker-friendly replace markers makes it a substantially greater pain in the ass for a regular human to edit the template to working order.

PR #680, opened 2017-06-12
Dockerize cytube and fixes for reverse proxying

davidsiaw:
Following a previous [PR that was on master](https://github.com/calzoneman/sync/pull/679), opening this branch to make changes to PR to upstream.

---

This is a set of changes to sync that I used to make dockerizing it possible and simplifying deployment.

It also has a few changes that fix some bugs that cropped up when reverse proxying https and wss requests to cytube (that is, allowing both ports 8080 and 1337 to be accessed through a server that only has port 80 and 443 accessible through a reverse proxy via separate domain names).

There is a change to the cytubefilters repository reference as well, this repository has a change that enables cytubefilters to be built on Alpine Linux (which is a small distro that I wanted to use as the base of my docker container).

The config template has been changed to enable one to simply run cat config.yaml.template | envsubst to generate a proper config file. This is far from complete but modifies some of the more important variables in the config file and hopefully will go some ways to improving the quality of life for people intending to deploy cytube.

In addition I have written a Dockerfile that will describe a build of cytube that should work out of the box and is self contained for testing, greatly shortening the time required to set up a test server.

I use this set of changes live on a private sync server I use with my friends.

I'm opening this pull request to open discussions about the changes and hopefully some of the changes will make it in.


README.md
```diff
@@ -80,6 +80,11 @@ Installing
 Installation instructions are available here: https://github.com/calzoneman/sync/wiki/CyTube-3.0-Installation-Guide
 
 
+Docker
+------
+
+`docker pull davidsiaw/sync:3.0`
```
Xaekai: Pull requests upstream shouldn't make dependencies of your fork.

config.template.yaml
```diff
@@ -4,11 +4,11 @@
 # user: username to authenticate as
 # password: password for user
 mysql:
-  server: 'localhost'
-  port: 3306
-  database: 'cytube3'
-  user: 'cytube3'
-  password: ''
+  server: '${MYSQL_HOST}'
```
Xaekai: Is there any way to make Docker use it's own separate config template? These changes make it much more annoying for regular humans to edit the template to working order

config.template.yaml
```diff
@@ -216,7 +217,7 @@ channel-blacklist: []
 # server-synched raw file playback.  This requires the following:
 #   * ffmpeg must be installed on the server
 ffmpeg:
-  enabled: false
+  enabled: true
```
Xaekai: I'm pretty sure this is disabled by default for a reason.

package.json
```diff
@@ -25,7 +25,7 @@
     "csrf": "^3.0.0",
     "cytube-common": "git://github.com/CyTube/cytube-common",
     "cytube-mediaquery": "git://github.com/CyTube/mediaquery",
-    "cytubefilters": "git://github.com/calzoneman/cytubefilters#67c7c69a",
+    "cytubefilters": "git://github.com/davidsiaw/cytubefilters",
```
Xaekai: If you've made changes to `cytubefilters` that are necessary to make this work, you should make a PR of those changes in that repo instead of changing this dependency.

run.sh
```diff
@@ -1,8 +1,64 @@
 #!/bin/sh
 
+#plagiarized from wangxian/alpine-mysql
```
Xaekai: This should be in it's own separate file for Docker as well because you're making a lot of assumptions about the user environment that the upstream script shouldn't make. For example that the user is even using MySQL since CyTube is targeting Knex.js now and the user could be using Postgres.

Xaekai (changes requested):

davidsiaw:
This build will fail for now because PCRE on calzoneman/cytubefilters has a bad config.h for Alpine Linux. Until that change is somehow incorporated, this Dockerfile is useless.

docker/docker-run.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+#plagiarized from wangxian/alpine-mysql
+if [ -d /app/mysql ]; then
+  echo "[i] MySQL directory already present, skipping creation"
+else
+  echo "[i] MySQL data directory not found, creating initial DBs"
+
+  mysql_install_db --user=root > /dev/null
+
+  if [ "$MYSQL_ROOT_PASSWORD" = "" ]; then
```
calzoneman, line 10: I would prefer to be "secure by default".  If the password is missing, instead of defaulting to something insecure, an error should be raised preventing the application from starting.

docker/docker-run.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+#plagiarized from wangxian/alpine-mysql
```
calzoneman, line 2: Is it compatible with their licensing to redistribute the install script like this?

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM alpine:3.6
+
+ADD . /app
+
+RUN cd app && sh docker/container-install.sh
+
+WORKDIR /app
+
+ENV MYSQL_HOST localhost
```
calzoneman, line 9: Doesn't it kind of defeat the purpose of docker to run a full stack inside one container?  The [MariaDB Docker image](https://hub.docker.com/_/mariadb/), for example, gives an example of how to link a database container with an application container.  There is also the case of connecting to an external database (e.g. AWS RDS, or a self-hosted db server), as I do today (3 app servers connecting to 1 db server).

calzoneman (commented):
A few remaining thoughts:

  * Does this support HTTPS over the socket.io socket?  The existing configuration method is a mess and I can't remember under what circumstances it generates secure connections if the application itself is running insecurely and being proxied.  As of right now, core only supports proxying HTTP requests; socket.io connections do not support proxying.
  * Do you want to be the maintainer of the Docker-related code?  I have plans to refactor several things in the future, including supporting databases other than MySQL/MariaDB, and refactoring the configuration.  My worry is that this will either create more work for my dev/test/release cycle to maintain the Docker image myself, or I will get lazy and the Docker image will fall behind/break.  I've thought about switching to Docker deployments one day, but right now all the live servers are on full VMs.
  * It looks like there is only one small patch to existing CyTube code in this PR; do you think it would make sense to centralize the Docker image in its own repository under the [CyTube](https://github.com/cytube) organization?

docker/docker-run.sh
davidsiaw: tbh I don't know. MIT is pretty permissive but I don't know for sure.

docker/docker-run.sh
davidsiaw: the idea here is that because the db isn't even accessible unless you are able to and do exec into the docker container, there is no way for you to connect to it, it does not matter.

Dockerfile
davidsiaw: My thinking is that for testing purposes it will have a db sitting inside the image. This means you can just run the container and cytube works. That said, you are right, it is always more elegant to have the image only contain the software and nothing else, and if you need a db you make another service (container or otherwise) and connect to it.

davidsiaw: The reason I did this was pretty selfish, I just wanted to get it working, and then make it better, so the next few iterations of the container on my fork is likely going to do away with the internal mysql server

davidsiaw:
In response to your thoughts:

1. It has always been my opinion (and my only my own) that a webservice concentrate on serving up content insecurely in a private network and be reverse proxied to by a public facing webserver that properly supports HTTPS, WSS and all the configuration that comes with that.

My current setup tells the container to be completely insecure and then I reverse proxy traffic into the appropriate ports with nginx containers. Works with the websockets just fine.

If you want I can show you how I do it. Right now its just a set of docker containers using nginx, docker-gen and letsencrypt and two DNS entries.

There is however one small bug that causes everyone who logs in to have the IP of the reverse proxy. I can put a fix in for that soon. (imagine global banning that IP LOL)

2. I can maintain the docker code. It should not be much work/code. I also use it myself. It is also easy to build/test/release the container. We can add this to the CI script.

3. Now that you say it it does make sense. My only concern is how we would trigger builds for that other repository on commits to this repo.

docker/docker-run.sh
calzoneman: We will need to check on that before this can be merged in.

Dockerfile
calzoneman: There is actually some value in having an all-in-one setup for the admins who just want to get a small server up and running and don't need the separation, I'm just wondering if there's a way to make it optional, or to split it up into multiple single-concern images that can optionally be orchestrated by a parent "all in one" package (I'm not super familiar with Docker administration).

If this PR is basically just a PoC/Alpha version that you intend to expand upon, then I'm fine with that, but it would be nice in the meantime to have a short readme that explains what exactly is installed.

calzoneman:
>It has always been my opinion (and my only my own) that a webservice concentrate on serving up content insecurely in a private network and be reverse proxied to by a public facing webserver that properly supports HTTPS, WSS and all the configuration that comes with that.

For smaller setups this is probably fine, but I found that with 1k+ users, proxying WSS had a nontrivial performance penalty (almost double CPU load, because the application is mostly IO-bound, and proxying a websocket amounts to just shuffling things around in network buffers in nginx).  I still proxy HTTP requests.

>Works with the websockets just fine.

Well, yeah, but as you've discovered, it doesn't respect any proxy headers like `X-Forwarded-For` ;)

>Now that you say it it does make sense. My only concern is how we would trigger builds for that other repository on commits to this repo.

Looks like Travis CI can't do cross-repo hooks natively (I didn't investigate whether it could with submodules), but there must be some script out there to do this.  I could also add an HTTP webhook to a simple endpoint that triggers a CI build.  But I'm not sure you would really want CI on the Docker image; I tend to push things that aren't fully tested and then shake them down on my beta host before deploying and considering it stable.

I've created https://github.com/CyTube/cytube-docker and added you as a collaborator there, if you'd like to continue the work there.

Dockerfile
davidsiaw: We can basically have an image that has no MySQL, but depending on the envvars set (say when MYSQL_HOST is localhost or 127.0.0.1) then the startup script will install and provision a local db and start cytube connected to it.

Yes it is very much a PoC, mostly to show that cytube can be dockerized and dockerizing it will save a huge amount of time otherwise spent on the gruelling installation process described in the wiki

davidsiaw:
All we want is for the docker image to be built and tested and then discarded to make sure the setup script still works in case a change changes the installation process.

Feel free to close the pull request and/or cherry pick any commits. I will continue the docker work on the cytube-docker repo

PR #682, opened 2017-06-15
make ACP use HTTPS only if it is enabled

davidsiaw:
This change really just clarifies things a bit for the server. Only generate a HTTPS link if you tell the server you are using HTTPS explicitly.

If you use HTTPS already websockets will refuse to be anything besides WSS.
If you use HTTP, websockets can use WSS but chances are you don't have that setup and it wouldn't work.

### DIFF
--- a/src/web/acp.js
+++ b/src/web/acp.js
@@ -28,7 +28,7 @@ function checkAdmin(cb) {
  */
 function handleAcp(req, res, user) {
     var sio;
-    if (req.secure || req.header("x-forwarded-proto") === "https") {
+    if ( Config.get("https.enabled") && (req.secure || req.header("x-forwarded-proto") === "https") ) {
         sio = Config.get("https.domain") + ":" + Config.get("https.default-port");
     } else {
         sio = Config.get("io.domain") + ":" + Config.get("io.default-port");


calzoneman:
Doesn't this introduce a bug/regression?  Suppose `Config.get("https.enabled")` is false, and the `X-Forwarded-Proto` header is `https`.  Then after this PR, the server will generate a non-secure link, which will not work because the `X-Forwarded-Proto` header is indicating the client is on a secure page, and active content security policy applies.

Arguably the existing code is incorrect, depending on your interpretation of how the service should behave if direct HTTPS is disabled but the reverse proxy terminates HTTPS connections (I consider this to be "undefined behavior" in the C sense of "it might work but I don't guarantee what it will do"), but this PR is incorrect in a different way.

This kind of conflicting behavior is one of the reasons I plan to move away from auto-generated links and require the administrator to explicitly specify which endpoints are to be used, putting the onus of understanding proxies and network topology on the administrator and not on the software.

davidsiaw:
> require the administrator to explicitly specify which endpoints are to be used

I would love that. Should we do that instead?

calzoneman:
That is the long term plan as part of simplifying the configuration process, however it will require a little more thought and planning because it's going to be a breaking change for all servers and require manual intervention from admins to reconfigure.

calzoneman:
I'm going to close this PR for the reasons discussed above.  Restructuring the configuration to remove these autogenerated edge cases is on the TODO list, and I think this particular problem will rarely be encountered.

PR #683, opened 2017-06-15
Make it possible to trust all proxies in config

davidsiaw:
It seems the WebConfig section for specifying trusted proxies is still hardcoded. This presents a problem because trusting proxies is the first step to recognizing the validity of X-Forwarded-For

There are times when one wants to trust all proxies that connect to cytube, i.e. in a completely private network where the only connections come through a reverse proxy.

This PR gives the ability to tell the server that.

### DIFF
--- a/config.template.yaml
+++ b/config.template.yaml
@@ -253,3 +253,9 @@ twitch-client-id: null
 
 poll:
   max-options: 50
+
+# Proxy trust settings 
+# set trust-all to true if you know anyone who is capable of connecting to your server is
+# trustworthy (for example if you proxy into a private network)
+proxy:
+  trust-all: false
--- a/src/io/ioserver.js
+++ b/src/io/ioserver.js
@@ -184,6 +184,7 @@ function ipForwardingMiddleware(webConfig) {
     }
 
     function isTrustedProxy(ip) {
+        if (Config.get("proxy.trust-all") === true) { return true; }
         return webConfig.getTrustedProxies().indexOf(ip) >= 0;
     }
 


davidsiaw:
Obviously the ideal fix is to enable one to set the trusted proxy IP addresses. But I don't know what the plan with the webConfig is at this time, and there seems to be references to oldConfig, possibly implying that the current configuration style is to be thrown away at some point.

Should that be the case this is more a temporary patch than anything else.

calzoneman:
Trust all seems a little dangerous.  It would be better to accept CIDR range syntax like [express does](http://expressjs.com/en/4x/api.html#trust.proxy.options.table) so you could whitelist the appropriate RFC 1918 private network block for your setup (or Cloudflare's IPs, if, heaven forbid, you attempt to get this working with Cloudflare).  Express supports this by default, but it should be possible to reuse the same module express depends on to achieve the same behavior in socket.io.  Is this something you'd like to look into?  It's been in the ultra-low-priority bucket of my TODO list for like 2 years, never had a reason to work on it before because I proxy locally on loopback.

>Obviously the ideal fix is to enable one to set the trusted proxy IP addresses. But I don't know what the plan with the webConfig is at this time, and there seems to be references to oldConfig, possibly implying that the current configuration style is to be thrown away at some point.

The idea behind `webConfig` was to abstract out all of the config access in an injected object so that I could migrate config formats seamlessly without intrusive changes to the code.  It's another relic of a project I never got around to finishing.  The references to `oldConfig` are how the `webConfig` object is populated from the legacy `config.yaml` map.

davidsiaw:
Changing it to allowing CIDR sounds good. If express does this it doesnt seem like it should be too much work

PR #684, opened 2017-06-16
Single source of truth for AFK

### LINKED ISSUE #678: Single Source Of Truth issue with AFK status
There are several members involved in this gang of chaos

But first, a picture says a thousand words.
image

When addUser handler builds the userlist object, it assigns a seperate "afk" property from the meta, and also assigns meta to a property directly. [line 509 callbacks.js](https://github.com/calzoneman/sync/blob/00a65a1584bf3e2dd78efb1a678829b935814ee0/www/js/callbacks.js#L509)

the setAFK handler only modifies the separate AFK property
the setUserMeta handler only modifies the real meta property

This can lead to the following situation:
User joins room. `afk` and `meta.afk` are both `false`
User goes afk. `afk` is true. `meta.afk` is `false`
User is then muted while afk. `afk` and `meta.afk` are both `true`
User voteskips. `afk` is `false`, `meta.afk` is `true`

I propose disposing of setAFK entirely, and leaving metadata in the meta property where it belongs. Anything that needs to access the AFK of a user can do so via `findUserlistItem(username).data().meta.afk`

Xaekai:
Resolves #678

### DIFF
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -87,7 +87,6 @@ function formatUserlistItem(div) {
         profile: div.data("profile") || { image: "", text: ""},
         leader: div.data("leader") || false,
         icon: div.data("icon") || false,
-        afk: div.data("afk") || false
     };
     var name = $(div.children()[1]);
     name.removeClass();
@@ -95,7 +94,7 @@ function formatUserlistItem(div) {
     name.addClass(getNameColor(data.rank));
     div.find(".profile-box").remove();
 
-    if (data.afk) {
+    if (div.data("meta") && div.data().meta.afk) {
         div.addClass("userlist_afk");
     } else {
         div.removeClass("userlist_afk");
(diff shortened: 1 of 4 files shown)


www/js/util.js
```diff
@@ -87,15 +87,14 @@ function formatUserlistItem(div) {
...
-        afk: div.data("afk") || false
     };
     var name = $(div.children()[1]);
     name.removeClass();
     name.css("font-style", "");
     name.addClass(getNameColor(data.rank));
     div.find(".profile-box").remove();
 
-    if (data.afk) {
+    if (div.data("meta") && div.data().meta.afk) {
```
calzoneman: Why the inconsistency of using `.data("meta")` and `.data().meta` in the same line?

Xaekai: was trying to be consistent actually, with the other lines in the block. I'm not sure why they are even pre-checking for the existence of `div.data("meta")`, not sure what situation would cause it to not exist

Xaekai:
Is that good?

www/js/util.js
calzoneman: If only JavaScript had a [null-safe accessor](https://github.com/davidyaha/ecmascript-optionals-proposal) :thinking:

PR #685, opened 2017-06-16
Customize channel path

### LINKED ISSUE #668: Customizable channel path
The channel path (currently `/r/`) should be a configurable option.

I'll do this next.

Xaekai:
Resolves #668 

I also reorganized the channel configs to be together to appease my own autism.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -70,7 +70,12 @@ var defaults = {
         "from-name": "CyTube Services"
     },
     "youtube-v3-key": "",
+    "channel-blacklist": [],
+    "channel-path": "r",
     "channel-save-interval": 5,
+    "channel-storage": {
+        type: "file"
+    },
     "max-channels-per-user": 5,
     "max-accounts-per-ip": 5,
     "guest-login-delay": 60,
@@ -102,7 +107,6 @@ var defaults = {
         "max-items": 4000,
         "update-interval": 5
     },
-    "channel-blacklist": [],
     ffmpeg: {
         enabled: false,
         "ffprobe-exec": "ffprobe"
@@ -113,9 +117,6 @@ var defaults = {
         "group": "users",
         "user": "nobody",
         "timeout": 15
-    },
-    "channel-storage": {
-        type: "file"
     },
     "service-socket": {
         enabled: false,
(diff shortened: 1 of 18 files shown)


www/js/ui.js
```diff
@@ -618,7 +618,7 @@ $("#shuffleplaylist").click(function() {
 /* load channel */
 
 var loc = document.location+"";
-var m = loc.match(/\/r\/([a-zA-Z0-9-_]+)/);
+var m = loc.match(/\/\w+\/([a-zA-Z0-9-_]+)/);
```
Xaekai: This is pretty disgusting code to begin with tbh.

1.) Why wasn't this wrapped in an IIFE?
2.) Whats wrong with `document.location.pathname.split('/')[2]` ?

calzoneman:
What is the allowable character set of `channel-path`?  The config loader should validate this with a regex and hard-crash with an explanation if it's invalid, otherwise someone's going to paste a URL in there or something and come running to IRC for help when express explodes cryptically due to a bad route.

`/r/` is also referenced here: https://github.com/calzoneman/sync/blob/3.0/www/js/callbacks.js#L440 but this shouldn't ever be used anymore (unless someone is manually sending a `login` frame instead of logging in through the website) and I don't think it even works because it doesn't include a CSRF token.

www/js/ui.js
calzoneman: That particular snippet dates back to [when the URL was first changed from `/?channel=whatever` to `/r/whatever`](https://github.com/calzoneman/sync/commit/3600a3114ab5f848ef8d3c260c2aec31bea7fbdf).  I don't know why it was done that way, it's probably not necessary (although you would still need to check that `document.location.pathname.split('/')[2]` is a valid channel name in your example).

Xaekai: Actually, I thought of a way to make all that totally obsolete. The channel template already knows the channels name. It can deposit it in the template and make life elegant.

Xaekai:
I presumed the same standards that applied to usernames and channel names would be suitable for this purpose. I will note them in the config template.

And yeah, I think we can just trash that cruft? All that's handled [here](https://github.com/calzoneman/sync/blob/3.0/templates/nav.pug#L46)

Xaekai:
Alright. I think that does it.

src/config.js
```diff
@@ -390,6 +391,12 @@ function preprocessConfig(cfg) {
     });
     cfg["channel-blacklist"] = tbl;
 
+    /* Check channel path */
+    if(!/^[-\w]+$/.test(cfg["channel-blacklist"])){
+        LOGGER.warn("Channel paths may only use the same characters as usernames and channel names.");
```
calzoneman: `LOGGER.error` and explicitly call `process.exit(1)` to prevent it from just being caught by the uncaught exception handler and chugging along.

calzoneman (commented):
Can you squash your commits?

Xaekai:
Squashed.

PR #687, opened 2017-06-17
requestUserlist

### LINKED ISSUE #681: Bots and userlist
Due to the different login path a bot takes, either the server needs to resend the userlist too when it sends a rank frame > 1 or the server needs a 'requestUserlist' listener

The problem is the bot has already been sent the userlist before it hits rank 2+ so only newly added users will have the `smuted`, `ip` and `aliases` properties.

Xaekai:
This resolves #681.
Additionally, clients ranked up 2+ after initial login will request a new playlist

### DIFF
--- a/src/channel/channel.js
+++ b/src/channel/channel.js
@@ -412,6 +412,8 @@ Channel.prototype.acceptUser = function (user) {
         self.modules[m].onUserPostJoin(user);
     });
 
+    // For bots who don't use cookies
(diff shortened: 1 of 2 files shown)


calzoneman:
I already fixed #681 with https://github.com/calzoneman/sync/commit/53cee986c6310070eae8ce076c5b405effe13201.  Is this needed for some other reason?

Xaekai:
No, that works.

PR #697, opened 2017-07-19
Video only button

0NDR:
Made a small change adding a Video-Only button, which makes the video player take up the entire browser window.

Because sometimes you just want to watch the stream without the chat, but dont want the window taking up an entire screen.

### DIFF
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -1845,8 +1845,7 @@ function chatOnly() {
     handleWindowResize();
 }
 
-function videoOnly()
-{
+function videoOnly() {
     var video = $("#videowrap").detach();
     video.prependTo($("body"));
     $("#wrap").remove();
@@ -1862,6 +1861,7 @@ function videoOnly()
     
     handleWindowResize();
 }
+
 function handleWindowResize() {
     if ($("body").hasClass("chatOnly")) {
         var h = $("body").outerHeight() - $("#chatline").outerHeight() -
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -1845,6 +1845,23 @@ function chatOnly() {
     handleWindowResize();
 }
 
+function videoOnly()
+{
+    var video = $("#videowrap").detach();
+    video.prependTo($("body"));
+    $("#wrap").remove();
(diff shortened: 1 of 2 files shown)


Xaekai:
* There is already a full screen button
* You didn't provide a way to restore without refresh

Xaekai:
image

calzoneman:
Fullscreen doesn't cover the case where someone may want to have the video windowed but without the chat.

0NDR:
>You didn't provide a way to restore without refresh

I was just mimicking the behavior of the "Chat-Only" button from the Layout drop down. I could work on a restore button, but then a restore button should probably also be added to the "Chat-Only" view.

Xaekai:
>someone may want to have the video windowed but without the chat

I suppose that's true. I withdraw my complaints.

www/js/util.js
```diff
@@ -1845,6 +1845,23 @@ function chatOnly() {
     handleWindowResize();
 }
 
+function videoOnly()
+{
```
calzoneman: Move brace to previous line for consistency with existing code style

www/js/util.js
```diff
@@ -1845,6 +1845,23 @@ function chatOnly() {
...
+    $("#chatwrap").remove();
+    video.css({
+        "min-height": "100%",
+        "min-width": "100%",
+        margin: "0",
+        padding: "0"
+    });
+    
+    handleWindowResize();
+}
```
calzoneman, line 1863: Add an extra newline between this function and the next one

calzoneman (changes requested):
Have you checked the browser console to check how many errors occur in video only mode?  I'd imagine it would be pretty noisy because the code makes countless assumptions about the state of the DOM (in particular, that all the things being removed in this PR will always be there).

It'd be nice if we could address this to clean it up a bit.  There is a long term goal to make the code more decoupled to allow for different layouts (chat only, video only, pop-out video, etc.), but that's not even reached the planning stages yet.

calzoneman:
Closing due to inactivity.  Feel free to rebase on top of the current HEAD and resubmit if you'd like.

PR #707, opened 2017-09-27
Upgrade nodemailer to 4.x

calzoneman:
* Upgrade nodemailer from old deprecated version to 4.x
  * Move email config to `conf/email.toml` and add additional options such as configurable template
  * Refactoring

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,20 @@
+2017-09-26
+==========
+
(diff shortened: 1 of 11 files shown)


Xaekai:
:sunbeam:

All hail Darude Calvin for progress on the transition to `toml` based configuration

PR #720, opened 2017-11-28
This resolves an issue where Google returns HTTP200 but provides an H…

### LINKED ISSUE #718: Detect and warn users about Google Drive login errors
Copying from #713 

>I think we are gonna have to start detecting if `accounts.google.com/ServiceLogin` is present in the response, and prompt the user in some fashion to log into their Gmail in another tab. This same problem presents on Chromium too.

Xaekai:
…TML redirect to a login portal instead of video data.

Closes #718

### DIFF
--- a/gdrive-userscript/cytube-google-drive.user.js
+++ b/gdrive-userscript/cytube-google-drive.user.js
@@ -8,7 +8,7 @@
 // @grant GM.xmlHttpRequest
 // @connect docs.google.com
 // @run-at document-end
-// @version 1.5.0
+// @version 1.6.0
 // ==/UserScript==
 
 try {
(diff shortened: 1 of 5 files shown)


gdrive-userscript/cytube-google-drive.user.js
```diff
@@ -206,15 +214,15 @@ try {
         'Violentmonkey' // https://github.com/calzoneman/sync/issues/713
     ];
 
-    function isRunningTampermonkey() {
+    function isTampermonkeyCompatible() {
```
calzoneman, line 217: In light of #721 I'm wondering if this shouldn't be inverted to "isn't greasemonkey" because GM is seemingly the only one so far that requires the sandboxed solution (which breaks Violentmonkey and might be responsible for that other one in #721 as well).

Not something you need to fix in this PR but getting the conversation started regardless.

www/js/data.js
```diff
@@ -1,4 +1,5 @@
 var CL_VERSION = 3.0;
+var GS_VERSION = 1.6; // Google Drive Userscript
```
calzoneman, line 2: I don't think there's much point in this being a float if it just gets converted into a string later.

gdrive-userscript/cytube-google-drive.user.js
Xaekai: so `isntShit()`

PR #740, opened 2018-03-30
Add Admin Setting - Block anonymous connections

OneEighteenNoTwo:
Hey. This was a feature that I thought might be nice to have and so I'm posting it here for consideration. I hope this is the right avenue for suggesting this sort of thing. Any feedback is most appreciated!

### DIFF
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -1,0 +1,28 @@
+"use strict";
+
+var ChannelModule = require("./module");
+var Flags = require("../flags");
+
+function AnonymousCheck(channel) {
+    ChannelModule.apply(this, arguments);
+}
+
+AnonymousCheck.prototype = Object.create(ChannelModule.prototype);
+
+AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
+    var chan = this.channel,
+    opts = this.channel.modules.options;
+   var anonymousBanned =  opts.get("block_anonymous_users");
+   if(anonymousBanned && user.isAnonymous()) {
+       user.socket.emit("errorMsg", { msg : "This channel has blocked anonymous users. Please provide a user name to join."});
+        user.waitFlag(Flags.U_LOGGED_IN, function () {
+            cb(null, ChannelModule.PASSTHROUGH);
+        });  
+        return;            
+    }
+    else{
+         cb(null, ChannelModule.PASSTHROUGH);
+    }
+};
+
+module.exports = AnonymousCheck;
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -12,6 +12,12 @@ AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
     opts = this.channel.modules.options;
    var anonymousBanned =  opts.get("block_anonymous_users");
    if(anonymousBanned && user.isAnonymous()) {
+        user.socket.on("disconnect", function () {
+            if (!user.is(Flags.U_IN_CHANNEL)) {
+                cb("User disconnected", ChannelModule.DENY);
+            }
+        });
+        
        user.socket.emit("errorMsg", { msg : "This channel has blocked anonymous users. Please provide a user name to join."});
         user.waitFlag(Flags.U_LOGGED_IN, function () {
             cb(null, ChannelModule.PASSTHROUGH);
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -11,6 +11,11 @@ AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
     var chan = this.channel,
     opts = this.channel.modules.options;
     var anonymousBanned =  opts.get("block_anonymous_users");
+    
+    if (user.socket.disconnected) {
+        return cb("User disconnected", ChannelModule.DENY);
+    }
+    
     if(anonymousBanned && user.isAnonymous()) {
         user.socket.on("disconnect", function () {
             if (!user.is(Flags.U_IN_CHANNEL)) {
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -1,5 +1,3 @@
-"use strict";
-
 var ChannelModule = require("./module");
 var Flags = require("../flags");
 
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -10,8 +10,8 @@ AnonymousCheck.prototype = Object.create(ChannelModule.prototype);
 AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
     var chan = this.channel,
     opts = this.channel.modules.options;
-   var anonymousBanned =  opts.get("block_anonymous_users");
-   if(anonymousBanned && user.isAnonymous()) {
+    var anonymousBanned =  opts.get("block_anonymous_users");
+    if(anonymousBanned && user.isAnonymous()) {
         user.socket.on("disconnect", function () {
             if (!user.is(Flags.U_IN_CHANNEL)) {
                 cb("User disconnected", ChannelModule.DENY);
--- a/src/channel/anonymouscheck.js
+++ b/src/channel/anonymouscheck.js
@@ -18,14 +18,14 @@ AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
             }
         });
         
-       user.socket.emit("errorMsg", { msg : "This channel has blocked anonymous users. Please provide a user name to join."});
+        user.socket.emit("errorMsg", { msg : "This channel has blocked anonymous users. Please provide a user name to join."});
         user.waitFlag(Flags.U_LOGGED_IN, function () {
             cb(null, ChannelModule.PASSTHROUGH);
         });  
         return;            
     }
     else{
-         cb(null, ChannelModule.PASSTHROUGH);
+        cb(null, ChannelModule.PASSTHROUGH);
     }
 };
 
--- a/src/channel/channel.js
+++ b/src/channel/channel.js
@@ -401,27 +401,6 @@ Channel.prototype.joinUser = function (user, data) {
     });
 };
 
-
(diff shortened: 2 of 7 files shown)


src/channel/channel.js
```diff
@@ -418,6 +418,14 @@ Channel.prototype.acceptUser = function (user) {
 
         this.logger.log("[login] Accepted connection from Tor exit at " +
                         user.displayip);
+    } 
+    else if(user.isAnonymous()){
+        var anonymousBanned = this.modules.options && this.modules.options.get("block_anonymous_users");
+        if (anonymousBanned) {
+            user.kick("This channel has banned anonymous users. Please register if you'd like to join.");
```
calzoneman: This would also block guests (unless they manage to type a guest name before the join callback fires) -- I feel like this should operate similar to the channel password feature and ask anonymous users to input a name before joining.

There might also be a case where people want to block guest names I guess, but I feel like that particular feature might be better solved by a whitelist (since registration doesn't require an email address or anything, "registered user" doesn't really have any more legitimacy than a guest name).

OneEighteenNoTwo:
I think there are probably three different feature possibilities here.  One feature, block non-registered accounts. Another feature, require guests to provide a guest name.  And finally, a third feature, having a white list.  

While they are different, I think maybe the first two do cover a lot of the same ground. I obviously liked the more broad blocking of all guests as a feature, but if you prefer blocking of guests with no name assigned, that's a really easy update.

Either way, I would say the white list definitely seems like a nice feature that would exist in addition to one, or both of those previously features.  While it could indirectly provide the feature of disallowing guests, I think it's appreciably different than either flavours of blocking guests in its scope.

src/channel/channel.js
OneEighteenNoTwo: Pardon the comment spam :) I realized I forgot to continue as a reply. 

I think there are probably three different feature possibilities here. One feature, block all guests. Another feature, block all guests without a guest name. And finally, a third feature, having a white list.

While they are different, I think maybe the first two do cover a lot of the same ground. I obviously liked the more broad blocking of all guests as a feature, but if you prefer blocking of guests with no name assigned, that's a really easy update. Edit: It's possible, too, these two ideas could co-exist, with the allowing of guests with names being an setting the channel owner could toggle.

Either way, I would say the white list definitely seems like a nice feature that would exist in addition to one, or both of those previously features. While a white list could provide the feature of disallowing guests, I think it's appreciably different than either flavours of blocking guests in its scope.

OneEighteenNoTwo: And just to avoid all risks of ideas getting twisted up because I'm using vocab that I'm not exactly rock solid on for this project - the intent of the original pull requests was to block all non-registered users from a channel, which I refereed to as anonymous users, even those who might have guest login named assigned.   In my reply to your response, when I'm referring to 'guests', I just mean any non-registered user.

calzoneman: In terms of the code, "anonymous" and "guest" are separate ranks with subtly different permissions; anonymous users have no name and hence do not appear in the userlist (unless they choose to enter a guest name later, in which case they become a guest).

Supporting blocking either one as options would be a possibility; I guess it depends on what your motivation is for using the feature.  Maybe I am missing some context on why the distinction between registered and guest name is important.  My goal here is to find the right balance between giving channel owners control of their channel and minimizing frustration and annoyance for users -- unless the channel has good reason to block guest names, I think "please enter a name to join" would be less annoying than "please register an account".

This also raises a question of whether channels which choose to enable this restriction should be disallowed from also setting the channel to "public" (i.e., list it on the index page).  It's a pretty annoying experience to browse the index page, click on something interesting, and immediately get a "You must register to continue" prompt.  It seems to me like this feature is mostly useful for private channels wishing to have a higher degree of accountability for who is joining.

Agree that whitelist is a separate feature.

OneEighteenNoTwo: Sweet, thanks for clearing up the lingo :) So, uh, I'm convinced by your arguments. There is a sort of tighter relationship to a user and a registered account, but like you said, the barrier to entry for registering is very low., so it's probably not that important  To that end, I do think disallowing anonymous connections, but allowing guest logins, satisfies what I would hope to have from the feature as a channel owner, which is just having more visibility with regards to who is in the channel.  I definitely agree that it's certainly a more user friendly experience for the anonymous / guest user, so that seems good, too.  Would I have your blessing to work that out?  

As far as my two cents regarding the question of whether or not to list channels publicly with this feature..to me, I think that only comes into play when considering the block guests + anonymous users feature,  In that case, the behaviour I would expect would be for these channels to not be listed for guest / anonymous accounts, but to be listed for registered users.  If the channel setting was just blocking anonymous users, but allowing guest logins, I think listing publicly could be O.K. provided we give the anonymous user an avenue to provide a guest name.

calzoneman: That seems reasonable.

src/channel/anonymouscheck.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+AnonymousCheck.prototype = Object.create(ChannelModule.prototype);
+
+AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
+    var chan = this.channel,
+    opts = this.channel.modules.options;
+   var anonymousBanned =  opts.get("block_anonymous_users");
+   if(anonymousBanned && user.isAnonymous()) {
+       user.socket.emit("errorMsg", { msg : "This channel has blocked anonymous users. Please provide a user name to join."});
```
calzoneman: If you click OK without thinking about it, this could get confusing because it says both "Not Connected" (on the userlist) and "Connected" in the chat, and in reality the user is half-connected (connected to the server, but not the channel, yet).

The password feature actually has this same issue, but in that case if you enter the wrong password, it immediately opens another prompt.  Maybe doing something similar here would be a short term solution until the UI is reworked for these cases.

src/channel/anonymouscheck.js
```diff
@@ -0,0 +1,26 @@
...
+function AnonymousCheck(channel) {
+    ChannelModule.apply(this, arguments);
+}
+
+AnonymousCheck.prototype = Object.create(ChannelModule.prototype);
+
+AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
+    var chan = this.channel,
+    opts = this.channel.modules.options;
+   var anonymousBanned =  opts.get("block_anonymous_users");
```
calzoneman: Inconsistent spacing

calzoneman (commented):
I remember there's some extra logic in accesscontrol about handling disconnects because if a user disconnects without entering the password, and the password module never calls back, the channel just gets stuck open.  That would happen here too.

src/channel/anonymouscheck.js
OneEighteenNoTwo: Well, if we would want to re-display the message anytime they clicked OK, I think instead of opening another prompt we might just want to remove the ability for the user to press O.K.  / dismiss the prompt.  How does that sound?  Alternatively, if there's a 'better' way to display a more persistent message in the channel in addition to the dialog, that seems like it would clear up any confusion.

calzoneman: What I meant was you could replace the OK prompt with a similar one to the one that the password feature uses, which requires you to enter a name to continue.

OneEighteenNoTwo: Right on. That's totally something I could rig up. The only concern I would have is do you think it might be a weird experience for users to have two inputs for guest login? That is, the one they're used to in the chat area and the dialog prompting them.

calzoneman:
I've been busy this week but I'll try to get this tested and merged tomorrow.

src/channel/anonymouscheck.js
calzoneman: It is a little weird, but I think the whole experience is already weird.  Ideally both this and the channel password case should block the whole UI until the user actually joins the channel (but I think that's a separate task).

src/channel/anonymouscheck.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+AnonymousCheck.prototype = Object.create(ChannelModule.prototype);
+
+AnonymousCheck.prototype.onUserPreJoin = function (user, data, cb) {
+    var chan = this.channel,
+    opts = this.channel.modules.options;
+    var anonymousBanned =  opts.get("block_anonymous_users");
+    if(anonymousBanned && user.isAnonymous()) {
+        user.socket.on("disconnect", function () {
```
calzoneman: The pre join check occurs asynchronously so the user might already be disconnected by the time you get here.  Look at the password module for how to check that.  After that, it appears to be working and I can merge it.

OneEighteenNoTwo: Ah ha, good catch. I think I got that cleared up in the latest commit.

src/channel/anonymouscheck.js
OneEighteenNoTwo: Ok, uh, do you think you could provide some acceptance criteria for what the ideal behaviour would be for this and channel password? I'd be down to take on the task if it's something that you think is worth adding.

calzoneman: I don't have a mockup or anything, but what I was imagining was that when the client received the socket frame asking for a password or guest login, it would cover the whole channel UI with a full page modal (except the navbar) and center the prompt to enter the required information.  I'm open to discussion on this, I just think that showing the entire UI as if it is connected, while the client is only half connected, is confusing.

OneEighteenNoTwo: Yeah, I totally agree with it being confusing with regards to the connected message in the channel area

Um, so I can't think of anything that seems bad about your suggestion here, which if I understand correctly is covering the channel UI to hide the connected message and providing the user with an input to enter whatever relevant data.  

The only alternative I can think of is keeping the behavior the way it is for the password / anonymous challenge, but instead of displaying the connected message in chat, displaying a message like 'Connection Pending' in the cases where the attempt to join the channel is prevented by a password or anonymous login. 

I'm not married to this alternative, and I'm not sure how uh, feasible it would be to implement off hand, but it was first alternate route that comes to mind.  Unless it seems better to you, or you have other ideas, I think your proposed approach is pretty solid, and that's something I would be able to implement - reviewing pending of course.

Issue #749, opened 2018-05-25
Initial userlist population quadratically sorts


Xaekai:
As brought to attention by Enzi on the Discord.

The list should be fully populated and then sorted once.

calzoneman:
The particular sort implementation probably also ought to be reviewed as it is 5 year old cruft.

calzoneman:
This has been deployed to https://beta.synchtube.me.

calzoneman:
This has been deployed live.

PR #750, opened 2018-05-25
Hoist sortUserlist outside of userlist population inner loop

calzoneman:
Mitigates #749.  After this change, `userlist` time is dominated by the DOM manipulation of `addUserToList` (optimizing all the other DOM manipulation is out of scope of this particular issue).


Xaekai:
Looks good to me.

PR #753, opened 2018-05-28
Show poll creator on hover of timestamp

### LINKED ISSUE #752: Poll timestamp should show poll creator on hover
Simple enhancement idea that's been bugging me for a minute now.

Xaekai:
Closes #752


calzoneman:
Regardless of my personal opinion about mobile browsers, people do still use them so we should probably avoid adding more hover elements to the UI.

Xaekai:
It's just a title attribute, this adds a feature for desktop and hurts mobile in no way.

PR #756, opened 2018-07-07
Add breaking spaces into footer

NPN:
On small mobile devices, the non-breaking spaces in the footer cause the links to overflow the container. (It's not so much of a problem with this version of the code, but the official website has an extra PayPal link which just pushes it over the edge.) This change replaces the non-breaking spaces after each middle dot with a normal space. This way, each middle dot stays with the link before it. (I chose this because I think it looks weird to have a line start with a dot.)

### DIFF
--- a/templates/footer.pug
+++ b/templates/footer.pug
@@ -2,7 +2,7 @@ mixin footer
   footer#footer
     .container
       p.text-muted.credit.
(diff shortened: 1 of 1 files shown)


calzoneman:
I merged this and deployed to https://beta.synchtube.me; can you confirm it works the way you expected?

NPN:
Yes, everything looks good.

Issue #767, opened 2018-09-01
[Feature Request] Support for HLS VoD Streams


Tomoyochan:
I am working on improving local playback capability and HLS VoD works partly.  The player plays the files however there is no synchronization of the video feed.  A public HLS VoD feed is here that shows the issue: https://content.jwplatform.com/manifests/yp34SRmf.m3u8

calzoneman:
Currently HLS support only targets livestreaming, for example in the custom media manifest format, it even verifies that HLS/DASH sources require `live:true`: https://github.com/calzoneman/sync/blob/3.0/src/custom-media.js#L32-L35.

I'd be curious to know what the value-add is of using HLS for prerecorded content instead of a static file (or set of files -- see https://github.com/calzoneman/sync/blob/3.0/docs/custom-media.md)

Tomoyochan:
It seems the detection failed as I can play both VoD files and it thinks that they are live streams. 

As for the value add, It would provide two items.
1) Multiple resolutions so that bandwidth is not wasted on the sending server.
2) Low bandwidth clients can get a lower bitrate feed.  This is the main reason that I have setup a HLS VoD environment for testing as some viewers of the static files are unable to watch as the BW of the file is to high.  I do not want to drop the quality for all viewers however.

The player in use also seemlessly switches between the qualities ala youtube so if a users connection is temporarily slow it will drop down temporarily.

I will read through the custom-media document.

calzoneman:
>It seems the detection failed as I can play both VoD files and it thinks that they are live streams.

To be clear: if you add an HLS file directly, it doesn't do any detection, it just assumes livestream (this predates the custom media manifest format).

Custom media manifests are intended to replace the self-hosted file support with a more advanced spec that allows users to define multiple formats/qualities for a single video as well as adding text tracks.

calzoneman:
Also: the status says "Experimental", but I really should change that.  It's officially supported.

Tomoyochan:
While functional the custom media manifest does not do things dynamically.

The users have to manually select the resolution they desire instead of the client choosing the best one for the conditions the way HLS does.   Further the client does not even default at the start to a quality that would be appropriate for the window size in my testing instead it chose the smallest one. 

To go back to why I would advise supporting HLS VoD:
1) It is a standard format used across a variety of services.
2) It is dynamic, youtube like setup, that adapts to the users window size and network quality automatically. 
[...]

PR #768, opened 2018-09-01
removed code that requires application/x-mpegURL to be live for custo…

Tomoyochan:
As per: [Feature Request] Support for HLS VoD Streams #767

I removed the lines that blocked the HLS stream from being added non-live. The only issue is that on my phone it shows a button that you have to click to start playback on android.  After clicking the button it does sync up as expected.

### DIFF
--- a/src/custom-media.js
+++ b/src/custom-media.js
@@ -30,8 +30,7 @@ const SOURCE_CONTENT_TYPES = new Set([
 ]);
 
 const LIVE_ONLY_CONTENT_TYPES = new Set([
-    'application/dash+xml',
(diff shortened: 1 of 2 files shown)


calzoneman:
Needs doc change, then we should be fine to merge.

calzoneman:
Hi, just checking in to see if you plan to include the change to custom-media.md removing the part about `application/x-mpegURL` being live-only on this PR.  Once that change is made, this looks good to merge.

@Tomoyochan

calzoneman:
Closing for inactivity.

PR #778, opened 2018-10-08
Chat notifications

really-need-an-api-key:
Adds a notification whenever a message is added to the chat.

### DIFF
--- a/templates/channel.pug
+++ b/templates/channel.pug
@@ -43,6 +43,7 @@ html(lang="en")
                 i#userlisttoggle.glyphicon.glyphicon-chevron-down.pull-left.pointer(title="Show/Hide Userlist")
                 span#usercount.pointer Not Connected
                 span#modflair.label.label-default.pull-right.pointer Name Color
+                span#notifications.label.label-default.pull-right.pointer Notifications
               #userlist
               #messagebuffer.linewrap
               input#chatline.form-control(type="text", maxlength="240", style="display: none")
--- a/templates/channel.pug
+++ b/templates/channel.pug
@@ -43,7 +43,6 @@ html(lang="en")
                 i#userlisttoggle.glyphicon.glyphicon-chevron-down.pull-left.pointer(title="Show/Hide Userlist")
                 span#usercount.pointer Not Connected
                 span#modflair.label.label-default.pull-right.pointer Name Color
-                span#notifications.label.label-default.pull-right.pointer Notifications
               #userlist
               #messagebuffer.linewrap
               input#chatline.form-control(type="text", maxlength="240", style="display: none")
--- a/templates/useroptions.pug
+++ b/templates/useroptions.pug
@@ -101,7 +101,7 @@ mixin us-chat
       +rcheckbox("us-sort-afk", "Sort AFKers to bottom")
       .col-sm-4
       .col-sm-8
-        p.text-info The following 2 options apply to how and when you will be notified if a new chat message is received while CyTube is not the active window.
+        p.text-info The following 3 options apply to how and when you will be notified if a new chat message is received while CyTube is not the active window.
       .form-group
         label.control-label.col-sm-4(for="#us-blink-title") Blink page title on new messages
         .col-sm-8
@@ -116,7 +116,13 @@ mixin us-chat
             option(value="never") Never
             option(value="onlyping") Only when I am mentioned or PMed
             option(value="always") Always
-      +rcheckbox("us-notifications", "Use notifications")
+      .form-group
+        label.control-label.col-sm-4(for="#us-notifications") Desktop notifications on new messages
+        .col-sm-8
+          select#us-notifications.form-control
+            option(value="never") Never
+            option(value="onlyping") Only when I am mentioned or PMed
+            option(value="always") Always
       +rcheckbox("us-sendbtn", "Add a send button to chat")
       +rcheckbox("us-no-emotes", "Disable chat emotes")
       +rcheckbox("us-strip-image", "Remove images from chat")
--- a/templates/useroptions.pug
+++ b/templates/useroptions.pug
@@ -116,6 +116,7 @@ mixin us-chat
             option(value="never") Never
             option(value="onlyping") Only when I am mentioned or PMed
             option(value="always") Always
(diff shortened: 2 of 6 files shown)


templates/useroptions.pug
```diff
@@ -116,6 +116,7 @@ mixin us-chat
             option(value="never") Never
             option(value="onlyping") Only when I am mentioned or PMed
             option(value="always") Always
+      +rcheckbox("us-notifications", "Use notifications")
```
calzoneman: I think desktop notifications should probably be handled the same way as the notification and page title options above -- it should be a tri-state (never / only when I am mentioned / always) instead of just on/off.

templates/channel.pug
```diff
@@ -43,6 +43,7 @@ html(lang="en")
                 i#userlisttoggle.glyphicon.glyphicon-chevron-down.pull-left.pointer(title="Show/Hide Userlist")
                 span#usercount.pointer Not Connected
                 span#modflair.label.label-default.pull-right.pointer Name Color
+                span#notifications.label.label-default.pull-right.pointer Notifications
```
calzoneman: Per my below comment about the setting, I think this can be removed since it creates confusion to configure the same setting in multiple places, and I think this setting is one that doesn't need to be quick-toggleable.

www/js/util.js
```diff
@@ -1649,6 +1672,10 @@ function addChatMessage(data) {
 
     pingMessage(isHighlight);
 
+    if (USEROPTS.notifications && document.hidden)
+    {
+        new Notification(data.username, {body: data.msg.replace(/<[^>]+>/g, ''), icon: null});
```
calzoneman: Since this is inside the chat message handler, you already have access to the div created for the message, so you can use `.text()`

calzoneman (changes requested):
Hi,

Thanks for your patience, I didn't have a chance to look at this until now.  Linking the previous/abandoned PR for desktop notifications for reference: https://github.com/calzoneman/sync/pull/633.

templates/useroptions.pug
really-need-an-api-key: Will do.

templates/channel.pug
really-need-an-api-key: My motivation for that was that I often leave the stream in the background, an and option to hide notifications was pretty handy when I was not actively participating in the chat and didn't particularly care about what was being written there at that moment.
Perhaps it should be a separate setting (mute all kinds/selected kinds of notifications, like a sort of do-not-disturb mode), or maybe it should cycle the setting through the three values from the comment above, or perhaps that's going too far...

www/js/util.js
really-need-an-api-key: Good point.

templates/channel.pug
calzoneman: This sounds like a pretty specific use case -- I don't have the numbers but I suspect a large portion of users never change the default notification settings at all (which only flash the title/play the notification sound on highlight/PM and not for every message).  In some channels, getting a notification for every message would be unusably spammy.

Anyways, there are a few reasons I'm not keen on adding this:

  * It gets cluttered (especially when channel scripts come along and add their own toggleables)
  * It confuses people -- shortly after adding the chat color toggleable I got a bunch of questions from people asking what it did, because clicking it has no immediately noticeable effect.  In the settings menu, settings have descriptions that inform users what they do, and it's clear it's a setting and not a button that will initiate some action.  Having a tri-state would exacerbate this problem.
  * Your original PR uses it to mean desktop notifications, but there are now actually 3 kinds of notifications (old fashioned blink tab title, notification sound, desktop notifications) and it's not clear which the toggleable would toggle

I think there is value in supporting desktop notifications as a core feature, but specifically the quick toggle part of it feels more appropriate as a channel/userscript level customization.

really-need-an-api-key: I see your point. Will remove the toggle.

By the way, chat messages that say "x joined" also get a highlight and a boop when "Only when I'm mentioned or PMed" is selected. Is that intended behavior?

www/js/util.js
```diff
@@ -1708,9 +1695,19 @@ function pingMessage(isHighlight) {
...
         }
+
+        if (USEROPTS.notifications === "always" || (USEROPTS.notifications === "onlyping" &&
+            isHighlight)) {
+            showDesktopNotification(notificationTitle, notificationBody.replace(/\[[^\]]+\]/g, ''));
+        }
     }
 }
 
+function showDesktopNotification(notificationTitle, notificationBody)
```
really-need-an-api-key, line 1706: Not going to lie here, I opted to introduce this function instead of just doing `new Notification()` in `pingMessage` because it will be easier to override in channel JS.

templates/channel.pug
calzoneman: The join message is a chat message, so it obeys the same notification policy.  That said, you shouldn't be receiving a join message for yourself -- so I think that should only occur if an IP with your name as an alias, or with your name as a substring of their name (due to that issue I've never fixed) joins.

www/js/util.js
```diff
@@ -1681,9 +1695,19 @@ function pingMessage(isHighlight) {
             isHighlight)) {
             CHATSOUND.play();
         }
+
+        if (USEROPTS.notifications === "always" || (USEROPTS.notifications === "onlyping" &&
+            isHighlight)) {
+            showDesktopNotification(notificationTitle, notificationBody.replace(/\[[^\]]+\]/g, ''));
```
calzoneman: What is this replacement for?

calzoneman: Ah, the message div is actually a wrapper that contains the username, the timestamp, and the message.  If you want only specific parts of that, you can pull those individual elements out and get the text out of those instead.

really-need-an-api-key: D'oh. Done here, and in PM callback.

templates/channel.pug
really-need-an-api-key: Oh that's probably it. I'm testing on localhost in a second private window.

calzoneman:
Closing via https://github.com/calzoneman/sync/commit/0bc866dbfaba0ef09dc703d6ef261bdf2e70b2ff

Issue #764, opened 2018-08-19
Self service conveniences


calzoneman:
A few things that people occasionally ask me to do manually that really ought to be self-service actions on the website:

  * [x] Delete an account (#464) -- landed in https://github.com/calzoneman/sync/commit/aa2348656dcf2fbd5f1562475ee4fab97e153816
  * [ ] View and clear aliases
  * [ ] Transfer a channel to a different owner

### Delete an account

Discussed in #464.  Needs to verify that the user doesn't own any channels first, and prompt the user to delete or transfer those channels before deleting the account.  Consider having a delayed deletion to prevent abuse.

### View and clear aliases

The account page should show which aliases are associated with the user's IP address and explain what aliases are and why the user may see aliases they don't own (VPNs, public networks, etc.).  The user should be able to clear their aliases, with some restriction on how often to prevent abuse.  Alternatively, consider just deprecating the aliases feature.

### Transfer a channel to a different owner

The channels page should allow the user to transfer the channel ownership to another user.  Initiating a transfer should not complete immediately, but prompt the other user to accept the transfer.  Once complete, the original owner is demoted to rank 4 and the channel's owner flag is set to the new owner.

Xaekai:
![image](https://user-images.githubusercontent.com/594255/49681989-102d2a80-fa60-11e8-93b2-a30ed5ea8eda.png)
First one done.

calzoneman:
Putting down some thoughts for `Transfer a channel to a different owner:`

  * Channel owner can request transfer to another user
  * The other user must accept the transfer before it goes through
  * Unaccepted transfers are abandoned after 7 days, or when the channel owner submits a new transfer request for the same channel
  * Users can see transfer requests from the my channels page, or maybe get an email notification about it

Issue #464, opened 2015-04-12
user deletion button


ghost:
I keep getting between 1-3 emails a week from people who want to delete their account. A button in the userlist in acp would be nice and/or a button on the profile tab or something for users to click themselves without me having to do it.

calzoneman:
I'm curious what the motivation is for deleting accounts to begin with.  The only identifiable account data is email address, password, and profile, all of which can easily be wiped by the user and the account just abandoned.  I'll go into detail about the technical issues tomorrow.

ghost:
I agree but users don't understand that's how user accounts work. Accounts
could also be set to be deleted after a set number of days without being
used like on irc services. Would prevent the database getting full of
unused accounts. Another idea could be to put in a link underneath the
change password/email page to the wiki with a page explaining why we don't
have a delete button.

calzoneman:
> Accounts could also be set to be deleted after a set number of days without being used like on irc services. Would prevent the database getting full of unused accounts.

Well if accounts were automatically deleted then it would make sense to add an interface to do it manually as well.  Unfortunately, deleting accounts presents some issues:
- It's inconvenient from a programming perspective, although the newer schema as of last summer makes it somewhat easier.  You'd need to delete information about the user (users table, channel_ranks table, user_playlists table), and decide what to do with other information associated with the user (channels the user owns, bans for that name).
- It opens up avenues for social engineering attacks.  Admittedly this is a stretch, but if a high profile user gets mad and deletes their account, someone could re-register it and trick people into modding the account again or something.

That said, I don't see any reason why, assuming the appropriate information is deleted from the database, anything would break as a result of deleting a user.

Another option is to add soft-deletion, that just flags an account as deleted, erases the profile/email information, and sets the password hash to NULL, thus rendering the account impossible to log into and removing any identifying information from the account.

Xaekai:
Ideas on this:
- Require the user to delete all their channels first.
- Mark the account as in "dormant state" which makes it unusable for 90 days
- Once the 90 days have passed delete everything.

This would prevent abuse of reuse of deleted accounts for social engineering via impersonation / ban evasions.

I also like the idea of autodeletion for seemingly abandoned accounts. Perhaps starting at 6 months for accounts that are barren of profile, email, and channels, with some sort of logic to make it longer for accounts that actually seem to have been used.

calzoneman:
I think I will actually pick this up.  My thought is:

  * Require all channels to be deleted or have ownership transferred first
[...]

PR #788, opened 2018-12-06
Account self delete

calzoneman:
As discussed in https://github.com/calzoneman/sync/issues/764 and https://github.com/calzoneman/sync/issues/464.

### DIFF
--- a/templates/account-delete.pug
+++ b/templates/account-delete.pug
@@ -1,0 +1,51 @@
+extends layout.pug
+
+block content
+  .col-lg-6.col-lg-offset-3.col-md-6.col-md-offset-3
+    if internalError
+      h2 Error
+      p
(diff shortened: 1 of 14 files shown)


templates/account-delete.pug
```diff
@@ -0,0 +1,51 @@
...
+    else if authFailed
+      h2 Authentication failed
+      p= reason
+    else if channelCount > 0
+      h2 Delete Account
+      p
+        | Your account cannot be deleted because you have one or more channels
+        | registered.  In order to delete your account, you must first&nbsp;
+        a(href="/account/channels") delete them
+        | &nbsp;or ask an administrator to transfer ownership of these channels
```
Xaekai, line 26: Next candidate for self-service: Relinquish channel ownership to another user

PR #816, opened 2019-06-09
add autoplay attribute to custom embed iframe tag, so autoplay works

mrx1983:
calzone said i should make a pull request. so here it is.
it would be great if 'allow="autoplay"' attribute is added to the generated iframe for custom embeds.
so autoplay works as expected.


calzoneman:
Merging, this will go out whenever the next deployment happens.

PR #839, opened 2019-12-05
update button labels

Olie440:


Xaekai:
There is no reason to change these strings.

Olie440:
I think it would make it a bit clearer as to what is happening.
I used what Spotify uses.

Olie440:
@calzoneman What do you think?

calzoneman:
I'm ok with making the button text more clear, but I'm not sure about the "Add to Playlist" label being used in this context; it could be confused with offering the ability to add the item to a saved/user playlist instead of the channel queue.

I think I would prefer if the buttons had a consistent verb, e.g. "Play next/Play last" or "Add next/Add last"; what do you think?

Xaekai:
It would seem the consistent verb is already `queue` internally. Why not `Queue next` and `Queue last`

Olie440:
@Xaekai Yeah sounds good!
Apologies for the way I came in, Last night I discovered the site and got excited about contributing to open source.

templates/channel.pug
```diff
@@ -111,9 +111,9 @@ html(lang="en")
                   .input-group
                     input#mediaurl.form-control(type="text", placeholder="Media URL")
                     span.input-group-btn
-                      button#queue_next.btn.btn-default Next
+                      button#queue_next.btn.btn-default Play Next
                     span.input-group-btn
-                      button#queue_end.btn.btn-default At End
+                      button#queue_end.btn.btn-default Add to Playlist
```
calzoneman: Looks like there is a second place that needs to be updated

Olie440: Good spot, updated

Olie440:
@calzoneman @Xaekai Btw if you have any tech debt you want doing let me know

Xaekai:
Can you squash your commits?

Olie440:
That should be a option on the merge button right?

Olie440:
@Xaekai Thanks so much!

PR #853, opened 2020-03-29
Fix channel password prompt

### LINKED ISSUE #852: Problematic z-index on div#needpw
Please fill out the templates below to the best of your ability, based on whether your problem is with using the website or with running your own server.

## Website Problem ##

**Please confirm whether you've tried the following debugging steps:**

  - [x] Clearing cache and refreshing the page (On Firefox, press Ctrl+F5.  On Chrome, press F12, then right click the refresh button and click "Empty Cache and Hard Reload")
  - [x] Disabling all browser extensions
  - [x] Using a clean channel with no customizations
  
### Description of the Problem ###

- What triggers the problem?
Opening a channel requiring a password on a small screen (e.g. phone).

- What happens?
The password prompt appears behind the guest login name box.

- What do you expect to happen instead?
The password prompt appears above the guest login name box.

The issue seems to be because the `div#needpw` element has an element style including `z-index: auto` overriding its CSS `z-index: 9999`.

Screen Shot 2020-03-29 at 9 33 32 pm

### System Information ###

  - **Operating System (Windows / Mac / Linux / Android / iOS):**
iOS / macOS (with responsive design mode)
  - **Web Browser (Firefox / Chrome / Other):**
Safari

Xaekai:
Add zin param to allow explicit z-index in JS generated dialog boxes.
Give needpass dialog explicit z-index

Resolves #852

### DIFF
--- a/www/js/util.js
+++ b/www/js/util.js
@@ -2097,11 +2097,20 @@ function waitUntilDefined(obj, key, fn) {
     fn();
 }
 
-function chatDialog(div) {
+/*
+    God I hate supporting IE11
+    https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
+    https://caniuse.com/#search=default%20function
+
+    This would be the ideal:
+    function chatDialog(div, zin = "auto") {
+*/
+function chatDialog(div, zin) {
+    if(!zin){ zin = 'auto'; }
     var parent = $("<div/>").addClass("profile-box")
         .css({
             padding: "10px",
-            "z-index": "auto",
+            "z-index": zin,
             position: "absolute"
         })
         .appendTo($("#chatwrap"));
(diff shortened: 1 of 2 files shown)


www/js/util.js
```diff
@@ -2097,11 +2097,20 @@ function waitUntilDefined(obj, key, fn) {
...
+    This would be the ideal:
+    function chatDialog(div, zin = "auto") {
+*/
+function chatDialog(div, zin) {
+    if(!zin){ zin = 'auto'; }
     var parent = $("<div/>").addClass("profile-box")
         .css({
             padding: "10px",
-            "z-index": "auto",
+            "z-index": zin,
```
calzoneman, line 2113: I'm kinda curious why it was originally set to auto explicitly.  Having the default here at least avoids breaking anything by introducing the 9999 to just this one case, but at some point we may need to look at the other places where chatDialog is used to see if there are similar bugs.

Xaekai: That was exactly my thinking, didn't know why it was like this and didn't want to break anything as a side effect.

PR #858, opened 2020-04-09
Replaced "display: none" check logic

Algoinde:
JQuery queries using getComputedStyle, which makes it impossible to change userlist behavior using CSS. This replaces the check with a direct style="" value check so the JS does not trip up if any CSS customizations to the list visibility were made.

Example: if `#userlist` is set to `display: block !important` in the custom CSS, the toggle will break and won't toggle between `none` and `block`. Implementing custom CSS leveraging this state change thus is not possible.

### DIFF
--- a/www/js/ui.js
+++ b/www/js/ui.js
@@ -817,7 +817,7 @@ $("#cs-emotes-import").click(function () {
 
 var toggleUserlist = function () {
     var direction = !USEROPTS.layout.match(/synchtube/) ? "glyphicon-chevron-right" : "glyphicon-chevron-left"
-    if ($("#userlist").css("display") === "none") {
+    if ($("#userlist")[0].style.display === "none") {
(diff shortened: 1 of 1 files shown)


Xaekai:
Strange, you made a feature branch, which is where you should have made this PR.

Yet you did [this](https://github.com/Algoinde/sync/commit/0e0593f6f10a0e5ed787f42ef000f44417fe0704). Kinda curious why? You can PR directly from a feature branch, then when PR is accepted you can just kill the feature branch and `git fetch upstream master` (though in this case of this repo replace master with 3.0)

Anyway, LGTM.

Algoinde:
This is the first time I am doing a pull request through GitHub interface, so I got lost a bit, sorry for that. I see what you mean now. Will this be fine if squashed?

Xaekai:
Here --> #859

PR #876, opened 2020-06-25
Send username in delete frames

deerfarce:
At the moment, you can't really see who's deleting items from the playlist without looking at the channel log. This adds a "username" property/key to delete frames which can be seen by those who have permission to delete videos (though I think that maybe everyone registered should be able to see it instead). Those without permission do not receive the username property, and server actions send [server], of course.
This way it's easier, for example, for bots to log exactly who deleted a certain video without pulling the full channel log.

### DIFF
--- a/src/channel/playlist.js
+++ b/src/channel/playlist.js
@@ -594,7 +594,7 @@ PlaylistModule.prototype.handleDelete = function (user, data) {
     var plitem = this.items.find(data);
     self.channel.refCounter.ref("PlaylistModule::handleDelete");
     this.semaphore.queue(function (lock) {
-        if (self._delete(data)) {
+        if (self._delete(data, user.getName())) {
             self.channel.logger.log("[playlist] " + user.getName() + " deleted " +
                                     plitem.media.title);
         }
@@ -694,7 +694,7 @@ PlaylistModule.prototype.handleJumpTo = function (user, data) {
         this.channel.logger.log("[playlist] " + user.getName() + " skipped " + title);
 
         if (old && old.temp && old !== to) {
-            this._delete(old.uid);
+            this._delete(old.uid, user.getName());
         }
     }
 };
@@ -874,7 +874,7 @@ PlaylistModule.prototype.handleUpdate = function (user, data) {
  * == Internal playlist manipulation ==
  */
 
-PlaylistModule.prototype._delete = function (uid) {
+PlaylistModule.prototype._delete = function (uid, username) {
     var self = this;
     var perms = this.channel.modules.permissions;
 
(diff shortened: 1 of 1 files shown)


src/channel/playlist.js
```diff
@@ -1046,7 +1052,7 @@ PlaylistModule.prototype._addItem = function (media, data, user, cb) {
         if (this.items.insertAfter(item, this.current.uid)) {
             if (existing && !allowDuplicates) {
                 item.temp = existing.temp;
-                this._delete(existing.uid);
+                this._delete(existing.uid, "[server]");
```
calzoneman, line 1055: I think it is weird to send a username in this case.

The socket frames sent by CyTube for playlist management are more like commands for the playlist UI rather than a log of actions initiated by some user.  In this particular case, the fact that it is implemented as a delete is an implementation detail -- logically what is happening is a move (this particular branch is when Queue Next is used on an item already on the playlist in a different position; assuming duplicates are not allowed, it moves the queued item to the front of the queue by inserting a new item there and removing the old one).

At various points in the past I have had ideas to improve the channel log to function more like a machine-readable audit log; I never got around to it before, but I feel like that would be a better solution to this type of use case (see who did what) by recording actions (and relevant session data) instead of the low level mechanics of how the UI should update in response to those actions.

src/channel/playlist.js
```diff
@@ -694,7 +694,7 @@ PlaylistModule.prototype.handleJumpTo = function (user, data) {
         this.channel.logger.log("[playlist] " + user.getName() + " skipped " + title);
 
         if (old && old.temp && old !== to) {
-            this._delete(old.uid);
+            this._delete(old.uid, user.getName());
```
calzoneman, line 697: Similar, I think it is weird to say that someone who skips from a temporary video to another one "deleted" the temporary video (this also applies below in the case where the server advances to the next item in the list which triggers the removal of the temporary item).

src/channel/playlist.js
```diff
@@ -594,7 +594,7 @@ PlaylistModule.prototype.handleDelete = function (user, data) {
     var plitem = this.items.find(data);
     self.channel.refCounter.ref("PlaylistModule::handleDelete");
     this.semaphore.queue(function (lock) {
-        if (self._delete(data)) {
+        if (self._delete(data, user.getName())) {
```
calzoneman, line 597: `getName()` could be null for anonymous users.  I think the voteskip log coalesces this to `(anonymous)`.

This actually makes me realize I think the log statement immediately below this is also flawed and will log "null deleted blahblah".

calzoneman:
Closing for inactivity.

PR #888, opened 2020-09-22
adjust sendUserMeta rank comparison

deerfarce:
Right now, those who use mute and shadowmute seem to miss out on the setUserMeta frame if their rank is the same as the minimum permission for muting users. This should fix it. Doesn't seem to break anything, but will now include users with a rank of -1 when -1 is passed as minrank for actions such as unmute -- just something to note in case that isn't intended for whatever reason.


deerfarce:
also, check my trips

PR #889, opened 2020-11-02
Support enabling custom media subtitle by default

animeavi:
If there's a cleaner way to do the CoffeeScript code instead of copying every attribute twice, let me know.
Tested locally and works as intended.

### DIFF
--- a/docs/custom-media.md
+++ b/docs/custom-media.md
@@ -119,6 +119,8 @@ Each text track entry is a JSON object with the following keys:
     [`text/vtt`](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API).
   * `name`: A name for the text track.  This is displayed in the menu for the
     viewer to select a text track.
+  * `default`: Enable track by default.  Optional boolean attribute to enable
+    a subtitle track to the user by default.
 
 **Important note regarding text tracks and CORS:**
 
@@ -148,7 +150,8 @@ for more information about setting this header.
         {
           "url": "https://example.com/subtitles.vtt",
           "contentType": "text/vtt",
-          "name": "English Subtitles"
+          "name": "English Subtitles",
+          "default": true
         }
       ]
     }
--- a/player/videojs.coffee
+++ b/player/videojs.coffee
@@ -96,12 +96,16 @@ window.VideoJSPlayer = class VideoJSPlayer extends Player
             if data.meta.textTracks
                 data.meta.textTracks.forEach((track) ->
                     label = track.name
-                    $('<track/>').attr(
+                    attrs =
                         src: track.url
                         kind: 'subtitles'
                         type: track.type
                         label: label
-                    ).appendTo(video)
+
+                    if track.default? and track.default
+                        attrs.default = ''
+
+                    $('<track/>').attr(attrs).appendTo(video)
                 )
 
             @player = videojs(video[0],
--- a/src/custom-media.js
+++ b/src/custom-media.js
@@ -209,6 +209,7 @@ function validateTextTracks(textTracks) {
     if (!Array.isArray(textTracks))
         throw new ValidationError('textTracks must be a list');
 
+    let default_count = 0;
     for (let track of textTracks) {
         if (typeof track.url !== 'string')
             throw new ValidationError('text track URL must be a string');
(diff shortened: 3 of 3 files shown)


docs/custom-media.md
```diff
@@ -119,6 +119,8 @@ Each text track entry is a JSON object with the following keys:
     [`text/vtt`](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API).
   * `name`: A name for the text track.  This is displayed in the menu for the
     viewer to select a text track.
+  * `default`: Enable track by default.  Optional boolean attribute to enable
```
calzoneman, line 122: The [validator](https://github.com/calzoneman/sync/blob/3.0/src/custom-media.js#L204) needs to be updated to validate this new field.

However, another thing to consider is that in the proposed change, `default` is an attribute of the text track descriptor itself; nothing inherently prohibits the user from specifying `default: true` on multiple text tracks.  This raises a question of what should happen in that scenario.  One option would be to add validation against it, but that doesn't really prevent the user from making the mistake in the first place.  Another option would be to have a separate top-level `defaultTextTrack` field that points to the URL of one of the `textTracks` entries, which also requires validation to ensure that the default track is actually present in the manifest.

player/videojs.coffee
```diff
@@ -96,12 +96,21 @@ window.VideoJSPlayer = class VideoJSPlayer extends Player
...
-                        type: track.type
-                        label: label
-                    ).appendTo(video)
+                    if track.default? and track.default
+                        $('<track/>').attr(
+                            src: track.url
+                            kind: 'subtitles'
+                            type: track.type
+                            label: label
+                            default: ''
```
calzoneman: How did you test this?  The PR title implies you want the video player to enable subtitles, but my reading of the [relevant MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTrackElement) is that this attribute merely specifies that among multiple track options, this one should be the default selected, and doesn't necessarily control whether the track ought to be enabled or not (especially since CyTube is using a videojs plugin for the track selector -- one would probably need to consult the docs of that particular plugin to know for certain).

Regardless, I feel that designation of a default track should be independent of the user's choice to enable text tracks or not.  One clean way to do this without introducing more UI clutter would be to do what YouTube does: when the user turns captions on, remember that the next time a YouTube video plays, and similarly when they turn it off.

Something worth considering as well is whether the manifest format ought to have a field on the text track for the language code (ISO or IETF standard) and let the user specify their preferred subtitle language instead of relying on the manifest author to choose an appropriate default.

player/videojs.coffee
```diff
@@ -96,12 +96,21 @@ window.VideoJSPlayer = class VideoJSPlayer extends Player
...
+                        ).appendTo(video)
+                    else
+                        $('<track/>').attr(
+                            src: track.url
```
calzoneman: You could probably avoid this if-else duplication by doing something like

```coffeescript
attrs =
  src: track.url
  # other attrs

if track.default? and track.default
  attrs.default = ''

$('<track/>').attr(attrs).appendTo(video)
```

calzoneman (changes requested):

animeavi:
My thinking with this was like, you're watching some French video with English softsubs, instead of telling everyone to enable the subtitles, you could do it by default, I only tested by setting one subtitle to default so far.
I'll see about making the changes and more testing in a bit.

animeavi:
Made most changes and added some basic validation.
For handling multiple defaults (if they exist) I was thinking of adding something like a `default_count` [here](https://github.com/calzoneman/sync/blob/dc8ac8a6a1e54751d26ff763c2e7de85dd7afe6f/player/videojs.coffee#L97) and only allowing the attribute to be set if it's `0`. What do you think?

From my tests, if multiple tracks are set to default, the topmost one gets priority.

calzoneman:
>My thinking with this was like, you're watching some French video with English softsubs, instead of telling everyone to enable the subtitles, you could do it by default

Hmm.  I generally prefer to keep the default experience very plain and allow users to opt-in to whatever they'd like.  I do think it would be reasonable to have the player remember that subtitles were enabled, as a convenience so that the user doesn't have to continually select it every time a new video is played.

However, if you're having to *tell* people to enable subtitles, perhaps the problem is really that the player doesn't make it obvious that subtitles are available.

>For handling multiple defaults (if they exist) I was thinking of adding something like a default_count here and only allowing the attribute to be set if it's 0. What do you think?

I would prefer if the server just prevented you from adding multiple defaults in the same manifest.

animeavi:
Well it can be done in the validator the same way, I'll look for a way that doesn't look too messy.

animeavi:
How does that look, @calzoneman?

PR #890, opened 2020-11-18
Add support for raw AV1/Opus

animeavi:


calzoneman:
Regarding AV1, as noted on the [MDN entry](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#AV1) support for this is relatively new and relatively experimental -- some browsers may not support it at all, and others may be only supporting it with slow/inefficient software decoders.  As a datapoint, my desktop browser only very recently added support for it, and my mobile browser cannot play AV1.

Thus far my position has been to only enable codecs that have very wide and well-established browser support in order to minimize the chance that a user will encounter unplayable videos on their device (and also to minimize the chance that they then ask the support channel for help).  I would be somewhat willing to reconsider this position with input from the community and also contingent on a prerequisite pull request to improve the error reporting mechanism for unsupported codecs to make it crystal clear that (1) the problem is caused by the user's browser not supporting the codec, (2) the decision to use that particular codec was explicitly made by the person who queued the video and not the website itself, and therefore (3) the website admins cannot provide support in this case.

Opus I don't have a problem with, since the compatibility table implies it has been widely supported for a while.

As an aside, any change to the ffprobe raw file support should also be made to the custom-media path, since that is really intended to be the successor feature.  Since you recently submitted a different PR to the custom-media path, I assume you know where to look for that.

animeavi:
I assume I don't need to change anything in `custom-media.js` for the AV1 (`video/mp4` and `video/webm` already cover it right?).
I did forget to add Opus there though.

calzoneman:
I forgot that the custom-media validator only specifies MIMEs/containers and not stream codecs.

animeavi:
Added opus in `custom-media.js`, your call if you wanna merge this.
Feel free to keep it open until you feel support is good enough.

calzoneman:
As of writing, the AV1 support on MDN states that Chrome, Edge, Firefox, and Opera all support AV1, and I've verified Firefox locally.  IE and Safari don't, but I'm not super concerned about that.  Merging.

PR #894, opened 2021-01-03
Block age-restricted yt videos from being queued

aleves64:
Needs https://github.com/CyTube/mediaquery/pull/12 to work.


src/channel/playlist.js
```diff
@@ -973,6 +973,10 @@ PlaylistModule.prototype._addItem = function (media, data, user, cb) {
         }
     }
 
+    if (media.meta.rating) {
```
calzoneman: Probably best to check if rating is equal to `"ytAgeRestricted"` in case there are other ratings that are not blocked.

https://developers.google.com/youtube/v3/docs/videos#contentDetails.contentRating.ytRating

src/get-info.js
```diff
@@ -71,6 +71,9 @@ var Getters = {
             if (video.meta.blocked) {
                 meta.restricted = video.meta.blocked;
             }
+            if (video.meta.ytRating) {
+                meta.rating = video.meta.ytRating;
```
calzoneman: Any particular reason to drop the `yt` prefix when cloning this property from the mediaquery model to the legacy cytube model?

calzoneman:
I'm working on publishing your mediaquery change as 0.0.25, it will need to be updated here: https://github.com/calzoneman/sync/blob/3.0/package.json#L13

(I never got around to cleaning up the npm publish vs. master branch distinction in that repo)

src/channel/playlist.js
aleves64: Added this. Initially thought I'd leave it as is since the current API says that's the only value it can have, but they might add new stuff there.

src/get-info.js
aleves64: Not really. Having it as ytRating is probably a little clearer.

calzoneman:
I edited your FAQ entry a little bit, but thanks for adding that.  LGTM.

calzoneman:
Deployed https://beta.synchtube.me for testing

PR #895, opened 2021-01-08
Fix '/youtube/v3/search was not found' in config sample

AlexandriaOL:
What it says on the tin. Fix the sample curl request so it doesn't error.

### DIFF
--- a/config.template.yaml
+++ b/config.template.yaml
@@ -116,7 +116,7 @@ io:
 # 5. Test your key (may take a few minutes to become active):
 #
 #    $ export YOUTUBE_API_KEY="your key here"
-#    $ curl "https://www.googleapis.com/youtube/v3/search?key=$YOUTUBE_API_KEY&part=id&maxResults=1&q=test+video&type=video"
+#    $ curl "https://youtube.googleapis.com/v3/search?key=$YOUTUBE_API_KEY&part=id&maxResults=1&q=test+video&type=video"
 youtube-v3-key: ''
 # Limit for the number of channels a user can register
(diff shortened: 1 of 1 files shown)


calzoneman:
The existing one works fine:

```
$ export YOUTUBE_API_KEY='<snip>'
$ curl "https://www.googleapis.com/youtube/v3/search?key=$YOUTUBE_API_KEY&part=id&maxResults=1&q=test+video&type=video"
{
  "kind": "youtube#searchListResponse",
  "etag": "xDGVEwA4PHypG4tDJboOkldp5dg",
  "nextPageToken": "CAEQAA",
  "regionCode": "US",
  "pageInfo": {
    "totalResults": 1000000,
    "resultsPerPage": 1
  },
  "items": [
    {
      "kind": "youtube#searchResult",
      "etag": "s5Kj5v_Hspi8H8OifxluVhM7tTY",
      "id": {
        "kind": "youtube#video",
        "videoId": "LXb3EKWsInQ"
      }
    }
  ]
}
```

If you received an error, perhaps your key is configured incorrectly.

calzoneman:
The [documentation](https://developers.google.com/youtube/v3/docs/search/list#http-request) also suggests that `https://www.googleapis.com/youtube/v3/search` is the correct URL.

screenshot-2021-01-08_19-54-31

AlexandriaOL:
Yeah, turns out there's a difference between getting the API key directly through `apis and services -> youtube v3 data api -> credentials` (which doesn't work) as opposed to `apis and services -> credentials` :/

PR #946, opened 2022-05-18
removed code that requires application/dash+xml to be live for custom media manifests

vt-idiot:
Tried to base these changes off #768 
I tested and can confirm that MPD VODs work just fine. Removing the live check will allow durations and automatic track progression. First commit to src/custom-media.js might have been _too_ much? I removed `LIVE_ONLY_CONTENT_TYPES` and the validation check entirely since the former would've been empty.

### DIFF
--- a/src/custom-media.js
+++ b/src/custom-media.js
@@ -30,10 +30,6 @@ const SOURCE_CONTENT_TYPES = new Set([
     'video/webm'
 ]);
 
-const LIVE_ONLY_CONTENT_TYPES = new Set([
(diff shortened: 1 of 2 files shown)


calzoneman:
Your PR caused a linter failure due to the data parameter no longer being used in validateSources.  I fixed it and pushed in https://github.com/calzoneman/sync/commit/dcfcee9a23af906492f9a11edd055976b244588a.

I didn't test this (well, asides from the unit tests), I took your word for it that it works.  YMMV.

PR #953, opened 2022-09-19
Dockerfile and docker support

davidsiaw:
I have recently needed to update my branch to the latest changes of the cytube branch and redid the Docker support for my own purposes.

I realize this PR is not immediately mergeable due to several issues:
1. cyzon may not want the Dockerfile to be in this repo
2. there are some invasive changes to the package json
3. it is not clear how to use the Dockerfile correctly (the README additions are very basic and assume a lot of things). I am open to improving the documentation for these purposes.

The changes are a lot simpler compared to when I last proposed docker support (5 years ago) due to advances in both Docker and the MySQL images that are now available, and improvements to cytube that no longer required the invasive changes I once needed to do to get it to work.

Unfortunately I have not thought deeply about how we can incorporate these changes as part of cytube-docker, which has been maintained by someone else for a while now. It is my opinion that these changes are much simpler being in the repository itself.

I am putting this PR in the hopes that perhaps someone will find it useful and perhaps someone will find this the key to providing more meaningful docker support in the future, and also to serve as a point of contact for anyone who wishes to ask me questions about how to set up cytube on Docker.

There are several things that can/need to be improved
1. Moving the startup to a non-root user internal to the container
2. Making the environment variables used to configure the server better
3. Automatically selecting cloudflare IPs instead of hardcoding them in the config template as it is right now
4. Automatically build a docker container and send it up to dockerhub. (this no longer requires giving perms to dockerhub for anything anymore)

It is my hope that we can find a path forward towards some level of rudimentary Docker support that does not need maintenance, as I did not have to do much to make this work and it is not very complicated to get it working.

In any case if cyzon does not wish for an open ended PR to be here he is free to close it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,26 @@
-FROM alpine:3.6
+FROM alpine
 
-ADD . /app
+ADD bin /app/bin
+ADD gdrive-userscript /app/gdrive-userscript
+ADD player /app/player
+ADD src /app/src
+ADD templates /app/templates
+ADD www /app/www
+ADD .eslintrc.js /app/.eslintrc.js
+ADD index.js /app/index.js
+ADD package.json /app/package.json
+ADD package-lock.json /app/package-lock.json
+ADD postinstall.sh /app/postinstall.sh
+ADD servcmd.sh.js /app/servcmd.sh.js
+ADD container-install.sh /app/container-install.sh
 
 RUN cd app && sh container-install.sh
-
-WORKDIR /app
 
 ENV MYSQL_HOST localhost
 ENV MYSQL_PORT 3306
 ENV MYSQL_DATABASE cytube
 ENV MYSQL_USER cytube
 ENV MYSQL_PASSWORD nico_best_girl
-ENV MYSQL_ROOT_PASSWORD ruby_best_girl
 ENV SYNC_TITLE Sync
 ENV SYNC_DESCRIPTION Sync Video
 ENV ROOT_URL http://localhost:8080
@@ -20,12 +29,13 @@ ENV IO_ROOT_URL http://localhost
 ENV IO_ROOT_PORT 1337
 ENV ROOT_DOMAIN localhost:8080
 ENV HTTPS_ENABLED false
-ENV TRUST_ALL_PROXIES false
+ENV COOKIE_SECRET aaa
+ENV IMMEDIATE_PROXY 172.16.0.0/12
 #ENV YOUTUBE_KEY
 #ENV TWITCH_CLIENT_ID
 
 EXPOSE 8080
-# EXPOSE 1337
+EXPOSE 1337
 # EXPOSE 8443
 
 ADD conf /app/conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM alpine:3.6
+
+ADD . /app
+
+RUN cd app && sh container-install.sh
+
+WORKDIR /app
+
+ENV MYSQL_HOST localhost
+ENV MYSQL_PORT 3306
+ENV MYSQL_DATABASE cytube
+ENV MYSQL_USER cytube
+ENV MYSQL_PASSWORD nico_best_girl
+ENV MYSQL_ROOT_PASSWORD ruby_best_girl
+ENV SYNC_TITLE Sync
+ENV SYNC_DESCRIPTION Sync Video
+ENV ROOT_URL http://localhost:8080
+ENV ROOT_PORT 8080
+ENV IO_ROOT_URL http://localhost
+ENV IO_ROOT_PORT 1337
+ENV ROOT_DOMAIN localhost:8080
+ENV HTTPS_ENABLED false
+ENV TRUST_ALL_PROXIES false
+#ENV YOUTUBE_KEY
+#ENV TWITCH_CLIENT_ID
+
+EXPOSE 8080
+# EXPOSE 1337
+# EXPOSE 8443
+
+ADD conf /app/conf
+ADD config.template.docker.yaml /app/config.template.yaml
+ADD run.sh /app/run.sh
+
+WORKDIR /app
+
+CMD ["sh", "run.sh"]
--- a/README.md
+++ b/README.md
@@ -37,6 +37,38 @@ General help with the software and the website is also available on the IRC
 channel at [irc.esper.net#cytube](http://webchat.esper.net/?channels=cytube)
 during US daytime hours.
 
+Docker
+------
+
+Example for using the dockerfile on this repo.
+
+```
+docker build -t sync .
+docker network create sync
+
+docker run -d --name sync-db \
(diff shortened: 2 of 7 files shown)


calzoneman:
If you wanted to try to get this into a mergeable state, I'm open to working with you on that review.

Regarding the cytube-docker repository, I'm not sure it's really maintained, the last commit is 3 years ago and I'm not sure anyone is using it.  I think the concept is similar, with the main difference being that that repo provides a Docker compose file to model the application + MariaDB stack, which probably makes it a bit easier for users to get started.

davidsiaw:
Thanks. I am cleaning up the sources now to make them easier to inspect, and the next steps I will take is to remove some of the changes that turn out to no longer be necessary such as my old modifications to the cytube filters.

My goal here is to introduce a Dockerfile that will produce a docker image that anyone who uses docker-based solutions can readily pull in, and that the Dockerfile will require very little maintenance, and that most relevant configuration for docker is exposed via environment variables.

I would like to hear any concerns you may have.

davidsiaw:
Alright, so this is more or less all that you need to launch cytube on docker. This and knowledge of how to start up a mysql container and expose your ports (basic docker server knowledge).

All that is needed now is possibly an example docker-compose.yml or some documentation on how to set it up for people who just want to paste some stuff onto their console.

My goal is to make it so one can create a container image

```
docker build -t cytube .
```

And then run it

```
docker run -d --name db --network cytube_network ... mariadb
docker run -d --network cytube_network -p 8080:8080 cytube
```

And have a cytube.

I can go on to write some of that documentation when I catch a breather.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ADD servcmd.sh.js /app/servcmd.sh.js
+ADD container-install.sh /app/container-install.sh
+
+RUN cd app && sh container-install.sh
+
+ENV MYSQL_HOST localhost
+ENV MYSQL_PORT 3306
+ENV MYSQL_DATABASE cytube
+ENV MYSQL_USER cytube
+ENV MYSQL_PASSWORD nico_best_girl
```
calzoneman, line 23: For things like passwords (and the root URL, cookie secret, etc.) where we expect the user to always need to customize it, I think it would be preferable to leave it unset by default so that users are forced to fill in the appropriate value (usually missing values present a more obvious error message than silently trying to use some default value).

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENV SYNC_TITLE Sync
+ENV SYNC_DESCRIPTION Sync Video
+ENV ROOT_URL http://localhost:8080
+ENV ROOT_PORT 8080
+ENV IO_ROOT_URL http://localhost
+ENV IO_ROOT_PORT 1337
+ENV ROOT_DOMAIN localhost:8080
+ENV HTTPS_ENABLED false
+ENV COOKIE_SECRET aaa
+ENV IMMEDIATE_PROXY 172.16.0.0/12
```
calzoneman, line 33: Is this documented Docker behavior or an implementation detail?  I also assume this would only do anything if you also had a reverse proxy container running (e.g. nginx) in front of cytube?

README.md
```diff
@@ -37,6 +37,38 @@ General help with the software and the website is also available on the IRC
...
+-e IO_ROOT_URL=https://cytube.my.domain \
+-e ROOT_DOMAIN=cytube.my.domain \
+-e VIRTUAL_HOST=cytube.my.domain \
+-e VIRTUAL_PORT=8080 \
+-e LETSENCRYPT_HOST=cytube.my.domain \
+-e YOUTUBE_KEY=abcdefg \
+--network sync sync
+```
+
+Feedback
```
calzoneman: What's this blank section for?

README.md
```diff
@@ -37,6 +37,38 @@ General help with the software and the website is also available on the IRC
 channel at [irc.esper.net#cytube](http://webchat.esper.net/?channels=cytube)
 during US daytime hours.
 
+Docker
```
calzoneman, line 40: I would maybe put the docker instructions in a separate file (e.g. docs/docker.md) and then add a link to that from the top of https://github.com/calzoneman/sync/wiki/CyTube-3.0-Installation-Guide in case people want to use docker instead of following the usual installation instructions.

container-install.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+apk update
+apk add build-base python3 git npm mysql mysql-client curl gettext ffmpeg
+npm install npm@latest -g
+npm install
```
calzoneman, line 6: I think it would be preferable to just run these as separate `RUN` commands in the Dockerfile so that you can take advantage of Docker's caching and only re-run what is necessary (e.g., if you modify the source files, then you probably need to re-run the npm commands, but it's not necessary to reinstall all the apk dependencies).

run.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/sh
 
+envsubst < config.template.yaml > config.yaml
+
+mysqld --user=root &
```
calzoneman: I thought the idea was to run mysql as a separate container and network them?

calzoneman (commented):
Sorry it took me a while to get back to you.

run.sh
davidsiaw: Oh yeah good catch that should not be there

Dockerfile
davidsiaw: These are placeholder values, since docker does not allow the ENV command to have empty values. The user should set these values using the `-e` flag when starting the container. Building the container should be a one-time thing that is done by a CI task

Dockerfile
davidsiaw: This is a docker implementation detail. `172.16.0.0/12` is a known value, though users can change it, its usually left to be that. https://docs.docker.com/network/network-tutorial-standalone/

README.md
davidsiaw: Its just a space between the contents and the feedback section.

container-install.sh
davidsiaw: I placed them in a script file as a group of commands that don't have to be changed all the time, and are just peculiar to cytube.

Having them as separate `RUN` commands is only useful if you are constantly re-building the image, like when you changed cytube to the point where we need to install something in a very specific way, and hence we need to re-run the Dockerfile multiple times.

(The image should be built by a Github action)

README.md
calzoneman: Sorry, I meant the Feedback section itself is blank (there is a heading but nothing under it).

container-install.sh
davidsiaw: >  just peculiar to cytube.

just required by cytube, not really peculiar or anything. They shouldn't have to change much if ever, unless we require a new dependency or something.

willjr880:
Hello, I didn't realise something was in the works already. I have created a docker container (with the expectation that you run it behind a reverse proxy.) So far it appears to be working as expected. 
https://github.com/willjr880/cytube_docker

Feel free to use to your liking.
Wil

PR #958, opened 2022-11-20
Add whitelisted channel logic

AshesOfOwls:
Add whitelisted channel logic

### DIFF
--- a/src/partition/partitionchannelindex.js
+++ b/src/partition/partitionchannelindex.js
@@ -6,6 +6,7 @@ const LOGGER = require('@calzoneman/jsli')('partitionchannelindex');
 var SERVER = null;
 const CACHE_REFRESH_INTERVAL = 30 * 1000;
 const CACHE_EXPIRE_DELAY = 40 * 1000;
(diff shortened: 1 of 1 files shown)


AshesOfOwls:
Whoops sorry

calzoneman:
Regardless of the mistaken PR, I'll go ahead and point out that this is likely not the file you actually want to update (unless you are using the multi-server support, which is undocumented and pretty much no one outside of the cytu.be server would need to use it anyways).

You probably meant to update the [local channel index](https://github.com/calzoneman/sync/blob/3.0/src/web/localchannelindex.js) instead.

calzoneman:
Although, if you were the person messaging me about how to keep channels loaded even if they have 0 viewers, you probably don't need to muck with the public index code, you can just remove [this](https://github.com/calzoneman/sync/blob/3.0/src/server.js#L295-L297) instead so that channels are never unloaded from memory (the channel index only shows those channels which are "loaded").

AshesOfOwls:
@calzoneman That was my buddy I am working on it with. Thank you for the insight we are gonna try to implement that now

PR #967, opened 2023-03-22
Add max chat message length config option

Kethsar:
Just thought it would be nice. I wanted to increase the max message length of an instance I stood up recently, so that it could be closer to IRC message length. Decided to refactor it into an option instead of just changing the hard coded values.

### DIFF
--- a/templates/channel.pug
+++ b/templates/channel.pug
@@ -46,7 +46,7 @@ html(lang="en")
               #userlist
               #messagebuffer.linewrap
               form(action="javascript:void(0)")
-                input#chatline.form-control(type="text", maxlength=`${maxMsgLen}`, style="display: none")
+                input#chatline.form-control(type="text", maxlength=maxMsgLen, style="display: none")
                 #guestlogin.input-group
                   span.input-group-addon Guest login
                   input#guestname.form-control(type="text", placeholder="Name")
--- a/templates/channel.pug
+++ b/templates/channel.pug
@@ -46,7 +46,7 @@ html(lang="en")
               #userlist
               #messagebuffer.linewrap
               form(action="javascript:void(0)")
-                input#chatline.form-control(type="text", maxlength="320", style="display: none")
+                input#chatline.form-control(type="text", maxlength=`${maxMsgLen}`, style="display: none")
                 #guestlogin.input-group
                   span.input-group-addon Guest login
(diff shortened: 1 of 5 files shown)


templates/channel.pug
```diff
@@ -46,7 +46,7 @@ html(lang="en")
               #userlist
               #messagebuffer.linewrap
               form(action="javascript:void(0)")
-                input#chatline.form-control(type="text", maxlength="320", style="display: none")
+                input#chatline.form-control(type="text", maxlength=`${maxMsgLen}`, style="display: none")
```
calzoneman: Since `maxMsgLen` is being passed as a template parameter, you should be able to just do `maxlength=maxMsgLen` and not need the extra `` `${}` ``

calzoneman (changes requested):
I'm not overall opposed to adding this, but there are some gotchas that aren't obvious here.  For example, there are at least two different places where chat filters restrict output to 1000 characters (mostly as a safety mechanism to prevent abuse) [1](https://github.com/calzoneman/cytubefilters/blob/master/src/jsfilterlist.cc#L99) [2](https://github.com/calzoneman/sync/blob/3.0/src/channel/filters.js#L24).

 in your specific use case you mentioned you were trying to align with IRC which is capped at 510 (not counting CRLF) so you'd never run into this, but it'd be a little weird to let someone set a chat length limit of say 2000, but then if the user happens to activate any chat filters suddenly it's capped at 1000.

templates/channel.pug
Kethsar: I didn't try, I wasn't sure if it would automatically place it as a string or not if I did that. Will check.

Kethsar: Neat, that does work. I figured it would place it as a number and break things.

Kethsar:
Right, I forgot about those. I had mucked around the filter code a couple of months ago when looking into something before, and it slipped my mind here.
I'm not sure I want to mess with those at this time. Would you be opposed to just having a max of 1000 characters set, and checking/setting it in `preprocessConfig()`? Or would you prefer I close this for now until I may or may not get around to figuring out a good way to integrate the changes with the hardcoded values in the filters?

calzoneman:
>Would you be opposed to just having a max of 1000 characters set, and checking/setting it in preprocessConfig()?

That would be fine.  The comment in config.yaml should probably also alert the administrator to this limitation.

Issue #980, opened 2024-04-10
Dailymotion videos no longer play


xRyozuo:
**Please confirm whether you've tried the following debugging steps:**

  - [X] Clearing cache and refreshing the page (On Firefox, press Ctrl+F5.  On Chrome, press F12, then right click the refresh button and click "Empty Cache and Hard Reload")
  - [X] Disabling all browser extensions
  - [x] Using a clean channel with no customizations
  (also tried on several different devices and browsers)
  
### Description of the Problem ###
Videos from dailymotion dont seem to play anymore (tried adding new random videos, nada), this started happening sometime this week. The channel had been running and working for a few months with no issues. Now theres a blank black box in its stead but the header does display the correct title of the video.

### System Information ###
CONFIRMED IT DOESNT WORK ON:
  - **Operating System (Windows / iOS):**
  - **Web Browser (Firefox / Safari / Edge):**
  - **Error Messages Displayed:** N/A
  - **Screenshot of JavaScript Console:**
 unreachable code after return statement
[all.js:2:23140](https://api.dmcdn.net/all.js)
downloadable font: Glyph bbox was incorrect (glyph ids 27 29 124 127 128 136 149 150 156 162 181 198 201) (font-family: "Glyphicons Halflings" style:normal weight:400 stretch:100 src index:1) source: https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/fonts/glyphicons-halflings-regular.woff
Connecting to {"url":"https://zip.cytu.be:8443/","secure":true} [callbacks.js:1256:13](https://cytu.be/js/callbacks.js)
This integration method is deprecated. We advise using Player Embeds https://faq.dailymotion.com/hc/en-us/articles/4411096679954-Integrate-your-Player-Embed

calzoneman:
Hi, thanks for reporting this.  I'm no longer actively maintaining this software so we'll have to see if someone else is willing to redo the dailymotion implementation as it looks like they changed it.

xRyozuo:
Oh shame! So out of curiosity, what would I need to learn to do this
myself? I depend on the channel to fall asleep so I’m motivated lol

Xaekai:
The documentation for DailyMotion player APIs is here:
https://developers.dailymotion.com/sdk/player-sdk/web/#methods

The page referenced in the console errors of CyTube's webclient when you play a DailyMotion video current is here:
https://developers.dailymotion.com/guides/getting-started-with-web-sdk/#player-library-script


[...]

PR #983, opened 2024-05-11
Add player integration code removed from the dailymotion js sdk

hdfssk:
The broken dailymotion player described in https://github.com/calzoneman/sync/issues/980 was caused by Dailymotion removing direct player creation support from their js SDK. They’d set up a method to remove it a while ago, which they turned on with their [commit 75b4102](https://github.com/dailymotion/dailymotion-sdk-js/commit/75b410229aa10467026a53cc5778fde66f1e9be5). The js SDK, documented [here](https://developers.dailymotion.com/sdk/platform-sdk/javascript/), is loaded from https://api.dmcdn.net/all.js by sync & other projects. The SDK’s github readme currently says:
 
> ⚠️ WARNING: We no longer support this SDK for Player integration on web environments. It can be used for interaction with the [Platform API](https://developers.dailymotion.com/api/platform-api/) only. The Player should be embedded using the new default Player integration methods, info can be found [here](https://developers.dailymotion.com/sdk/player-sdk/web).

The new integration method they point to there seems to demand [making a dailymotion account, configuring a custom player w/ a unique ID, and loading their scripts based on that ID](https://developers.dailymotion.com/guides/getting-started-with-web-sdk/) . This is a can of worms, and it's not clear how this could work for sync in general.

Fortunately, it’s still possible to create a Dailymotion player with their js SDK’s object interface, instead of the functional interface that vanished when they decided to “Sunset player API v2". This PR does so to get Dailymotion sync support working again.

### DIFF
--- a/player/dailymotion.coffee
+++ b/player/dailymotion.coffee
@@ -18,7 +18,12 @@ window.DailymotionPlayer = class DailymotionPlayer extends Player
             if quality != 'auto'
                 params.quality = quality
 
-            @dm = DM.player('ytapiplayer',
+            @element = DM.$('ytapiplayer')
+            if not @element or @element.nodeType != Node.ELEMENT_NODE
+                throw new Error("Invalid player element in DailymotionPlayer(), requires an existing HTML element: " + @element)
+            if DM.Player._INSTANCES[@element.id] != undefined
+                @element = DM.Player.destroy(@element.id)
+            @dm = DM.Player.create(@element,
                 video: data.id
                 width: parseInt(VWIDTH, 10)
                 height: parseInt(VHEIGHT, 10)


player/dailymotion.coffee
```diff
@@ -18,7 +18,12 @@ window.DailymotionPlayer = class DailymotionPlayer extends Player
             if quality != 'auto'
                 params.quality = quality
 
-            @dm = DM.player('ytapiplayer',
+            @element = DM.$('ytapiplayer')
+            if not @element or @element.nodeType != Node.ELEMENT_NODE
+                throw new Error("Invalid player element in DailymotionPlayer(), requires an existing HTML element: " + @element)
```
calzoneman, line 23: Under what conditions could this happen?  If this could realistically happen to a user it'd be better to add a warning banner on the page rather than throw an error from the script in which case the user probably won't realize what happened.

Xaekai:
> The new integration method they point to there seems to demand [making a dailymotion account, configuring a custom player w/ a unique ID, and loading their scripts based on that ID](https://developers.dailymotion.com/guides/getting-started-with-web-sdk/)

This is precisely why I threw my hands up in the air. This is the same reason I haven't added Rumble support. Their APIs seem geared to allowing publishers to customize embeds on their own sites and the idea of an unrelated 3rd party using their embeds is persona non grata. I appreciate your efforts to subvert their enshittifications.

player/dailymotion.coffee
hdfssk: This is just the validation code from [dailymotion’s deprecated DM.player function](https://github.com/dailymotion/dailymotion-sdk-js/blob/43376ce9a1fe6375a53e4252f1f0fd021d9ca54d/src/core/player.js#L34). I don't think it's a realistic end user situation, it made more sense as a warning to outside programmers of the dailymotion js sdk that they'd passed in a bad parameter. In sync, it’s only useful for warning about a potential future programming error, and can safely be removed if you’d rather.
Shot May 12 2024, 1꞉39꞉17pm

hdfssk (commented):
oops, [left this in Pending](https://github.com/orgs/community/discussions/10369)

Xaekai:
Closes #980

PR #988, opened 2024-10-25
Fix/emote replace

esnyder232:
Replaced augint urls with static.coolhole.org urls.


esnyder232:
why does github default the base with calzoneman's?

